PR #4, opened 2022-07-21
Upgrade/remove vulnerable packages in test runner

itsrainy:
- Remove ncc as a dependency for the test runner itself; it's only needed as a dev dependency and is already specified as such.
- Run `npm audit fix` to fix 1 other vulnerability
- Run `npm build` to rebuild `compute-test-worker.js`


dgryski (approved):
LGTM

PR #7, opened 2022-09-15
Make the request/response limits configurable

andersjanmyr:
```
	fsthttp.Limits.SetMaxHeaderValueLen(32 * 1024)
```

### DIFF
--- a/fsthttp/limits.go
+++ b/fsthttp/limits.go
@@ -1,0 +1,49 @@
+package fsthttp
+
+import "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+
+// Limits handles HTTP limits
+var Limits = limits{}
+
+type limits struct {
+}
+
+// GetMaxHeaderNameLen gets the header name limit
+func (limits) GetMaxHeaderNameLen() int {
+	return fastly.MaxHeaderNameLen
+}
+
+// SetMaxHeaderNameLen sets the header name limit
+func (limits) SetMaxHeaderNameLen(len int) {
+	fastly.MaxHeaderNameLen = len
+}
+
+// GetMaxHeaderValueLen gets the header value limit
+func (limits) GetMaxHeaderValueLen() int {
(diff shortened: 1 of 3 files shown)


fsthttp/limits.go
```diff
@@ -0,0 +1,49 @@
...
+import "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+
+// Limits handles HTTP limits
+var Limits = limits{}
+
+type limits struct {
+}
+
+// GetMaxHeaderNameLen gets the header name limit
+func (limits) GetMaxHeaderNameLen() int {
```
dgryski, line 12: Effective Go recommends not having getter function names start with `Get`.  This should jus be `MaxHeaderNameLen()`.

dgryski (changes requested):
A small comment on function naming, but otherwise this looks great.  Thanks!

fsthttp/limits.go
andersjanmyr: Yes, I agree, it reads better. :)

fsthttp/limits.go
```diff
@@ -0,0 +1,49 @@
+package fsthttp
+
+import "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+
+// Limits handles HTTP limits
+var Limits = limits{}
+
+type limits struct {
+}
```
dkegel-fastly, line 9: I'd put the limits in the struct, that's four fewer globals...

andersjanmyr: It would create a circular dependency between the `fsthttp` and `internal/abi/fastly` packages.

dkegel-fastly: emoji-homer-hedge :-)

dgryski (approved):
LGTM

dgryski:
I'm going to see if I can tie the limits to a particular Request object.  That would solve having large allocations for every request (if it's not needed), and also prevent any data races from modifying these global variables from different goroutines.

PR #10, opened 2022-10-05
all: gofmt

dgryski:
This PR updates the comment formatting with the new (1.19) `gofmt`.


dgryski:
These are (for the most part) being hit by https://tip.golang.org/doc/comment#code .  This is easy enough to fix.  Generally, we just need to adjust *where* gofmt thinks the code block starts.

dgryski:
PTAL

joeshaw (approved):
👍

PR #11, opened 2022-10-06
add fsthttp.Transport, which is an http.RoundTripper implementation

joeshaw:
This allows existing Go code which uses an http.Client to be adapted to
work on Compute@Edge.

This is intended only to adapt existing code.  Using http.Client pulls
in substantially more code than is needed for a simple request,
resulting in slower compile times and larger binaries.  We recommend new
code use the fsthttp.Request type and its Send() method directly
whenever possible.

### DIFF
--- a/fsthttp/transport.go
+++ b/fsthttp/transport.go
@@ -1,0 +1,92 @@
+package fsthttp
+
+import (
+	"net/http"
+	"strings"
+)
+
+// Transport is an http.RoundTripper implementation for backend requests
+// on Compute@Edge.
+//
+// Compute@Edge requests must be made to a pre-configured named backend.
+// A default backend is set when the transport is created, but
+// additional backends can be added with the AddBackend method.
+//
+// This is primarily intended to adapt existing code which uses
+// configurable http.Client instances to work on Compute@Edge.  Using an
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM.

joeshaw:
@dgryski PTAL

fsthttp/transport.go
```diff
@@ -0,0 +1,97 @@
...
+	return &Transport{
+		defaultBackend: backend,
+		backends:       make(map[string]string),
+	}
+}
+
+// AddHostBackend adds a new host-to-backend mapping.  Multiple hosts
+// may be mapped to the same backend.
+func (t *Transport) AddHostBackend(host, backend string) {
+	t.backends[strings.ToLower(host)] = backend
```
dgryski, line 47: Do we want to protect `t.backends` with a mutex?

dgryski (approved):
LGTM.  Mutex addition is optional.

fsthttp/transport.go
joeshaw: I don't have a strong opinion on this.  It seems unlikely to me that users would add mappings concurrently, so my vote is to wait until it becomes a problem in practice.

PR #12, opened 2022-10-19
geo: add simple json fuzzer

dgryski:


dgryski:
We now use `encoding/json` for the `geo` package.

PR #13, opened 2022-11-02
object store integration

dgryski:


internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2048,3 +2048,119 @@ func GeoLookup(ip net.IP) ([]byte, error) {
...
+	).toError(); err != nil {
+		return "", err
+	}
+
+	b, err := io.ReadAll(&body)
+	if err != nil {
+		return "", err
+	}
+
+	return string(b), nil
```
JakeChampion, line 2124: @dgryski Am I correct that this is reading the Object Store value into memory and then parsing it into a string? If that’s the case then I think it won’t be possible to lookup large objects as they could exceed the allowed max memory on C@E. If possible, it'd be nice to have an option which returns a stream or iterator

dgryski: Hmm.. good point.  I was going for a simple API, but I'll switch this out a returning an `io.Reader` instead to allow streaming.

joeshaw: The same thing applies to `Insert`.

dgryski: @joeshaw Yeah, I was doing to do both sides.

dgryski:
Updated the API to be stream-based to match the hostcalls.

dgryski:
I'm not sure if having the stream-based API as the *only* one makes sense.  From a developer ergonomics point of view, I think people are going to want ones that return plain strings.

JakeChampion:
> I'm not sure if having the stream-based API as the _only_ one makes sense. From a developer ergonomics point of view, I think people are going to want ones that return plain strings.

Agreed, in the JS SDK we have an ObjectStore class and an ObjectStoreEntry class for the values contained in a store.
The ObjectStoreEntry class enables the application to convert to text/arraybuffer/json or to stream the entry.

```ts
declare class ObjectStore {
    constructor(name: string);
    lookup(key: string): Promise<ObjectStoreEntry | null>;
    put(key: string, value: BodyInit): Promise<undefined>;
}

declare class ObjectStoreEntry {
  get body(): ReadableStream;
  get bodyUsed(): boolean;
  text(): Promise<string>;
  json(): Promise<object>;
  arrayBuffer(): Promise<ArrayBuffer>;
  // And eventually formData and blob once we support them on Request and Response, too.
}
```

maybe something similar could be done for the go sdk?

dgryski:
Yeah, it's easy enough to return a type that implements both `io.Reader` (for streaming) and `Stringer` (for a `String()` method to return a simple `string`).  I'll see what that looks like...

objectstore/objectstore.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Store represents a Fastly object store
+type Store struct {
+	objectstore *fastly.ObjectStore
+}
+
+// Open returns a handle to the named object store
+func Open(name string) (*Store, error) {
+	o, err := fastly.OpenObjectStore(name)
```
JakeChampion, line 40: Non-Blocking Comment:
We could provide more specific errors here if we wanted to:
- If the name is empty
- If the name is more than 255 characters
- If the name contains control characters
- If the requested store does not exist

objectstore/objectstore.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &Store{objectstore: o}, nil
+}
+
+// Lookup fetches a key from the associated object store.
+func (s *Store) Lookup(key string) (*Entry, error) {
+	val, err := s.objectstore.Lookup(key)
```
JakeChampion, line 50: Non-Blocking Comment:
We could provide more specific errors here if we wanted to:

- If the key is empty
- If the key is more than 1024 characters
- If the key contains one or more of these characters`\n\r[]*?#`
- If the key starts with `.well-known/acme-challenge/`
- If the key is exactly the string `.` or the string `..`

objectstore/objectstore.go
```diff
@@ -0,0 +1,58 @@
...
+	return &Entry{Reader: val}, err
+}
+
+// Insert adds a key to the associated object store.
+func (s *Store) Insert(key string, value io.Reader) error {
+	err := s.objectstore.Insert(key, value)
```
JakeChampion, line 56: Non-Blocking Comment:
We could provide more specific errors here if we wanted to:

- All the same comments about the key for `Lookup`

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2048,3 +2048,114 @@ func GeoLookup(ip net.IP) ([]byte, error) {
...
+		return err
+	}
+
+	if _, err := io.Copy(body, value); err != nil {
+		return err
+	}
+
+	if err := fastlyObjectStoreInsert(
+		o.h,
+		prim.NewReadBufferFromString(key).Wstring(),
```
JakeChampion, line 2154: Non-Blocking Comment:
Does this mean we read the entire value into memory? If so, we might crash if the application is trying to send a HTTP response body into Object Store, as the application could exceed the max allowed RAM on C@E

_examples/objectstore/main.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2022 Fastly, Inc.
+
+package main
```
JakeChampion, line 3: Non-Blocking Comment:
It's great we have an example app and build it.

Should we also use these for end-to-end testing of the SDK? We do that for the JS SDK and it has helped increase confidence in our implementation and also caught differences between Viceroy and C@E

JakeChampion (approved):

PR #18, opened 2022-11-18
fsthttp: don't call Close() on abireq if we haven't written all the b…

dgryski:
…ytes

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -282,10 +282,15 @@ func (req *Request) Send(ctx context.Context, backend string) (*Response, error)
 
 	if shouldCopy {
 		go func() {
-			_, err := io.Copy(abiReqBody, req.Body)
-			errc <- maybeWrap(err, "copy body")
+			_, copyErr := io.Copy(abiReqBody, req.Body)
+			errc <- maybeWrap(copyErr, "copy body")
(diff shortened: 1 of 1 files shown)


joeshaw (approved):
👍

fsthttp/request.go
```diff
@@ -282,10 +282,15 @@ func (req *Request) Send(ctx context.Context, backend string) (*Response, error)
...
+			_, copyErr := io.Copy(abiReqBody, req.Body)
+			errc <- maybeWrap(copyErr, "copy body")
 			errc <- maybeWrap(req.Body.Close(), "close user body")
-			errc <- maybeWrap(abiReqBody.Close(), "close request body")
+			if copyErr != nil {
+				// If there was an error copying the body, we *don't* want to Close() the abi req.
+				// This tells the wasm server that the body is incomplete so it knows not to
+				// terminate the sent chunked body with a valid final chunk.
+				errc <- maybeWrap(abiReqBody.Close(), "close request body")
+			}
```
peterbourgon, line 293: Then shouldn't this be `if copyErr == nil`?

dgryski: Bah.  Yes.  Of course it should.

dgryski: https://github.com/fastly/compute-sdk-go/pull/20

PR #19, opened 2022-11-30
add Secret Store implementation

joeshaw:
This adds hostcall and SDK support for the Fastly Secret Store.  An
example is included for usage but the basic idea is:

```go
st, err := secretstore.Open("example_secretstore")
if err != nil {
    panic(err)
}

s, err := st.Get("my_secret")
if err != nil {
    panic(err)
}

pt, err := s.Plaintext()
if err != nil {
    panic(err)
}

fmt.Fprintf(w, "secret value: %q", v)
```

### DIFF
--- a/.github/actions/bootstrap-integration-test/action.yml
+++ b/.github/actions/bootstrap-integration-test/action.yml
@@ -4,21 +4,21 @@ runs:
   using: "composite"
   steps:
     - run: |
-        echo "Install Fastly CLI 3.0.0"
-        wget https://github.com/fastly/cli/releases/download/v3.0.0/fastly_3.0.0_linux_amd64.deb
-        sudo apt install ./fastly_3.0.0_linux_amd64.deb
+        echo "Install Fastly CLI 5.0.0"
+        wget https://github.com/fastly/cli/releases/download/v5.0.0/fastly_5.0.0_linux_amd64.deb
+        sudo apt install ./fastly_5.0.0_linux_amd64.deb
       shell: "bash"
     - run: |
-        echo "Install Viceroy 0.2.14..."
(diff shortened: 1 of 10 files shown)


joeshaw:
Still to do: Viceroy doesn't have support for these hostcalls yet, so the example is expected to fail on it.  Looks like we need an integration test for it as well (which I would expect to fail for now).

_examples/secret/main.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/fastly/compute-sdk-go/secretstore"
+)
+
+func main() {
+	fsthttp.ServeFunc(func(ctx context.Context, w fsthttp.ResponseWriter, r *fsthttp.Request) {
+		st, err := secretstore.Open("example_secretstore")
+		if err != nil {
+			fsthttp.Error(w, err.Error(), fsthttp.StatusBadGateway)
+			return
+		}
```
awilliams-fastly, line 17: What do you think about somehow showcasing `secretstore.ErrSecretStoreNotFound` and `secretstore.ErrSecretNotFound`? Could be something like this:

```suggestion
		switch {
		case errors.Is(err, secretstore.ErrSecretStoreNotFound):
			fsthttp.Error(w, err.Error(), fsthttp.StatusNotFound)
			return
		case err != nil:
			fsthttp.Error(w, err.Error(), fsthttp.StatusBadGateway)
			return
		}
```

joeshaw: Yeah, great idea.

joeshaw: Done in baed1f8

joeshaw:
Viceroy PR: https://github.com/fastly/Viceroy/pull/210

dgryski (approved):
LGTM

PR #22, opened 2023-01-20
objectstore: allow differentiating between empty and missing keys

dgryski:


internal/abi/fastly/types.go
```diff
@@ -214,6 +215,10 @@ const (
 //	(typename $body_handle (handle))
 type bodyHandle handle
 
+const (
+	invalidBodyHandle = bodyHandle(math.MaxUint32 - 1)
```
dgryski, line 219: In theory could add one of these for each of the different handle types.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2116,6 +2116,12 @@ func (o *ObjectStore) Lookup(key string) (io.Reader, error) {
 		return nil, err
 	}
 
+	// Didn't get a valid handle back.  This means there was no key
+	// with that name.  Report this to the caller by returning `None`.
+	if body.h == invalidBodyHandle {
```
dgryski, line 2121: Seems like a reasonable pattern we might want to repeat elsewhere.

dgryski:
I probably should update the object store example demonstrating checking for `NotFound`.

objectstore/objectstore.go
```diff
@@ -1,11 +1,14 @@
 package objectstore
 
 import (
+	"errors"
 	"io"
 
 	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
 )
 
+var ErrNotFound = errors.New("objectstore: key not found")
```
joeshaw, line 10: What does `Open` return if a store can't be found?  Should it be unified with this value?  Or should this value be called `ErrKeyNotFound` instead?

JakeChampion (approved):
Nice 😄 It's great that we will be able to differentiate between empty and missing 👍

joeshaw (commented):
I agree that the example would be good to update, as would a comment about returning `ErrNotFound` to the `Lookup` docs.

objectstore/objectstore.go
dgryski: I believe you get a `FastlyError` (forwarded from upstream) with `Status`
https://github.com/fastly/compute-sdk-go/blob/main/internal/abi/fastly/types.go#L46

joeshaw: A bigger question is maybe: should we be exposing `FastlyError` out of `internal` at all, or should we capture and have module-specific errors?

dgryski: Yeah, I was wondering about that.  I'll open a bug.

objectstore/objectstore.go
```diff
@@ -45,9 +48,20 @@ func Open(name string) (*Store, error) {
...
-// Lookup fetches a key from the associated object store.
+// Lookup fetches a key from the associated object store.  If the key does not
+// exist, Lookup returns the sentinel error ErrKeyNotFound.
 func (s *Store) Lookup(key string) (*Entry, error) {
 	val, err := s.objectstore.Lookup(key)
+	if err != nil {
+
+		// turn FastlyStatusNone into NotFound
+		if status, ok := fastly.IsFastlyError(err); ok && status == fastly.FastlyStatusNone {
+			return nil, ErrKeyNotFound
```
dgryski, line 59: I realize now that returning `nil` for the `Entry` changes behaviour for keys that don't exist.  Previously we would simply return an empty body (which was fine to read from), but. now we're returning a nil pointer than will segfault if you don't check for it.  I guess what's more likely to happen is that people will error out in places they wouldn't before, since most people are likely checking the error return, rather than having people *not* checking the error and blindly proceeding and now we're going to `panic` with a `nil` interface read.

dgryski:
CI failed first time, passed on re-run.  One of the tests has a data-race with output ordering.

joeshaw (approved):
👍

PR #24, opened 2023-01-23
Rename edgedict -> configstore.

dgryski:
This includes deprecation notices for the types and methods in edgedict, as well as semgrep patterns to update to the new API.


_integration_tests/fixtures/configstore/fastly.toml
```diff
@@ -23,6 +23,6 @@ service_id = ""
 
   [local_server.dictionaries]
 
-    [local_server.dictionaries.edge_dictionary]
-      file = "./_integration_tests/fixtures/edge_dictionary/edge_dictionary.json"
+    [local_server.dictionaries.configstore]
```
dgryski, line 26: viceroy still calls them `dictionaries` I guess.

configstore/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 Fastly, Inc.
+
+// Package configstore provides a read-only interface to config stores. Refer
+// to https://developer.fastly.com/reference/api/dictionaries/ for more
```
dgryski, line 4: No new page to link to for "config stores"?

joeshaw: https://developer.fastly.com/reference/api/services/resources/config-store/, though this might change.

joeshaw (approved):
One doc link update below, but otherwise LGTM

semgrep/edgedict.yml
```diff
@@ -0,0 +1,49 @@
+rules:
```
joeshaw, line 1: Might be good to include some docs or a makefile rule or something to show how to use these.  Or include them in CI.

dgryski: Oh right.  I'll include a README explaining that these are.

dgryski:
PTAL

fgsch:
Instead of rename can we keep the edgedict around and add a deprecated notice to it?

dgryski:
@fgsch Yes, that's exactly what this PR does.  It leaves a deprecated stub in `edgedict` that calls out to the `configstore` package as the base implementation, as well as adding deprecation notices to anything using `edgedict`.  It also uses a type alias for `edgedict.Dictionary` pointing to `configstore.Store`, and has a set of semgrep rules to handle the migration automatically (except for import rewriting).

JakeChampion (approved):

fgsch:
@dgryski doh, I should have read this more carefully! Sorry for the noise.

PR #26, opened 2023-02-09
Rework limits

fgsch:
I'm not sure about the interface but opening it for discussion.

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -13,6 +13,14 @@ import (
 	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
 )
 
+// RequestLimits are the limits for the components of an HTTP request.
+var RequestLimits = Limits{
+	maxHeaderNameLen:  fastly.DefaultMaxHeaderNameLen,
+	maxHeaderValueLen: fastly.DefaultMaxHeaderValueLen,
+	maxMethodLen:      fastly.DefaultMaxMethodLen,
+	maxURLLen:         fastly.DefaultMaxURLLen,
+}
+
 // Request represents an HTTP request received by this server from a requesting
 // client, or to be sent from this server during this execution. Some fields
 // only have meaning in one context or the other.
@@ -116,12 +124,12 @@ func newClientRequest() (*Request, error) {
 		return nil, fmt.Errorf("get client request and body: %w", err)
 	}
 
-	method, err := abiReq.GetMethod()
+	method, err := abiReq.GetMethod(RequestLimits.maxMethodLen)
 	if err != nil {
 		return nil, fmt.Errorf("get method: %w", err)
 	}
 
-	uri, err := abiReq.GetURI()
+	uri, err := abiReq.GetURI(RequestLimits.maxURLLen)
 	if err != nil {
 		return nil, fmt.Errorf("get URI: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


fsthttp/request.go
```diff
@@ -116,12 +124,12 @@ func newClientRequest() (*Request, error) {
 		return nil, fmt.Errorf("get client request and body: %w", err)
 	}
 
-	method, err := abiReq.GetMethod()
+	method, err := abiReq.GetMethod(RequestLimits.maxMethodLen)
```
joeshaw, line 127: I wonder if we should put request and/or response `Limits` fields inside the `Request` struct, and pull from that instead.

We could still keep the global `RequestLimits` and `ResponseLimits` values, and a newly-constructed `Request` would have its fields simply be a copy of the global ones, but it'd also give you the ability to make adjustments per-request.

The main upside to this is that if you know one particular backend request needs an especially big buffer you can avoid allocating it unnecessarily for all the other requests too.

internal/abi/fastly/types.go
```diff
@@ -278,23 +278,19 @@ const (
...
-	MaxHeaderValueLen = 8192
-	// MaxMethodLen is the method limit
-	MaxMethodLen = 1024
-	// MaxURLLen is the url limit
-	MaxURLLen = 1024
-	// InitialSecretLen is the initial size of the buffer to use for decrypting secrets
-	InitialSecretLen = 1024
-)
+// InitialSecretLen is the initial size of the buffer to use for decrypting secrets
+var InitialSecretLen = 1024
```
joeshaw, line 282: This is my fault, but I don't see any particular reason why this needs to be user adjustable.  I suggest making this `const initialSecretLen = 1024`.

joeshaw (commented):
My only big piece of feedback is that it'd be nice if we could set these limits per-request rather than having them be only globally set.  But other than that the details look good to me.

internal/abi/fastly/types.go
fgsch: Sure, I will update it.

fsthttp/request.go
fgsch: I'm happy to explore other venues but currently we prefetch everything before handling the incoming request to the caller, and it is at this point that we need to tweak the various limits.   For outgoing requests (created by `NewRequest`), they don't apply AFAICT.

fgsch:
The table below describe the current limits, the previous ones and what are the limits in the Rusk SDK:

| operation / version | Go 0.1.1 | Go 0.1.2 | Rust 0.9.1 (for reference) | changed in Go 0.1.2? |
|-|-|-|-|-|
| all header names (per header) | 10240 |  8192 | 8192 | Yes (lower) |
| all header values (per header) | 10240 | 8192 | 8192 | Yes (lower) |
| original header names (per header) | 10240 | 8192 | 8192 | Yes (lower) |
| single header value (unused in Go) | 65536 | 8192 | 8192 | Yes (lower) |
| method | 1024 | 1024 | 8192 | No |
| url | 10240 | 1024 | 8192 | Yes (lower) |
| tls cipher name | 10240 | 16384 | 128 | Yes (higher) |
| tls protocol | 10240 | 16384 | 32 | Yes (higher) |
| tls client hello | 10240 | 16384 | 512 | Yes (higher) |

internal/abi/fastly/types.go
fgsch: Done in https://github.com/fastly/compute-sdk-go/pull/26/commits/847263430e2b8842b5e66f3b3156709002bb4384

fsthttp/request.go
fgsch: Related to this, maybe it'd be better to name these `GlobalRequestLimits` and `GlobalResponseLimits` should we add per-requests ones in the future? Thoughts?

andersjanmyr:
It looks good to me. I'm happy as long as I can tweak the sizes when needed.

joeshaw (approved):
LGTM 👍

PR #28, opened 2023-03-06
use latest staticcheck in validate GH action

joeshaw:


dgryski (approved):
LGTM

dgryski:
Merging because my own PR is failing because of this.

joeshaw:
Unfortunately this is insufficient: `nilness` is still breaking.  Should we switch pulling from `master` or drop its check entirely?

dgryski:
I don't see a problem with pulling `nilness` from `master`.  I don't want to drop the check because it catches subtle bugs and has virtually no false positives.

joeshaw:
Ok.  Opened https://github.com/fastly/compute-sdk-go/pull/30 for that.

PR #30, opened 2023-03-08
lint: use nilness from master

joeshaw:
The latest release of nilness panics when used with Go 1.20.  See
golang/go#58296.  This was fixed, but hasn't been released yet, which
is why we pull from master.


dgryski (approved):
LGTM

dgryski:
Merging.

PR #33, opened 2023-03-26
Deprecate fstctx

dgryski:


fgsch:
This LGTM but only works for Tinygo >= 0.26 IIRC.
Should we update the CLI to mention 0.26 or higher?

fgsch (approved):

Integralist:
👋🏻 

If this is a `0.26` requirement, then let's try and coordinate the release so to avoid confusing users. 

We just need a CLI PR opened to update the following files...

- https://github.com/fastly/cli/blob/main/.fastly/config.toml#L8
- https://github.com/fastly/cli/blob/main/pkg/commands/compute/build_test.go

I can handle cutting the new-release side of things (as I've some other stuff I want to get out this week).

fgsch:
As per tinygo changelog, it was fixed in 0.26.0:

dgryski:
@fgsch Ah, excellent.  Thanks for looking that up.  Yes, we definitely want people using as recent a tinygo as possible.  Pushing them to 0.26 is no bad thing.

dgryski:
@Integralist Let me know when I can hit Merge then.

Integralist:
@dgryski will do. I'll try and get a release out tomorrow 👍🏻

Integralist:
@dgryski I've bumped the Go version constraint to >= 1.18 and Tinygo version constraint to >=0.26.0 and cut [a new Fastly CLI release](https://github.com/fastly/cli/releases/tag/v8.2.1).

That said I've just realised you've not removed `fstctx` from the SDK only 'deprecated' it so it still exists for users on an older version of the CLI who have a project with `fstctx`. So this PR should be fine to merge.

PR #34, opened 2023-04-11
rename object store to kv store

dennismartensson:
we are renaming the object store to kv store.


fgsch:
Should we deprecate it for a version or two as to not break existing code?

_examples/objectstore/main.go
dgryski, line 1: This enclosing directory should be renamed.

_examples/objectstore/main.go
```diff
@@ -8,12 +8,12 @@ import (
 	"strings"
 
 	"github.com/fastly/compute-sdk-go/fsthttp"
-	"github.com/fastly/compute-sdk-go/objectstore"
+	kvstore "github.com/fastly/compute-sdk-go/kvstore"
```
dgryski, line 11: Don't need a named import here -- package name and import name match.

dennismartensson:
We have agreed that sdk changes for this name change can be a breaking version update.

dgryski (changes requested):
Please a forwarding stub package with a deprecation notice like I did for the `edgedict` rename.  Writing `semgrep` rules for the transition is optional.

dgryski:
Do we need to release his SDK right away for this change?

dennismartensson:
let me give it a go and see what I end up with 😄

dgryski (approved):
LGTM.

kvstore/kvstore.go
```diff
@@ -0,0 +1,72 @@
...
+	b, err := io.ReadAll(e)
+	if err != nil {
+		return ""
+	}
+
+	e.validString = true
+
+	return string(b)
+
+}
```
peterbourgon, line 34:
```suggestion
	e.s = string(b)
	e.validString = true
	return e.s
}
```

kvstore/kvstore.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var ErrKeyNotFound = errors.New("kvstore: key not found")
+
+type Entry struct {
+	io.Reader
+
+	validString bool
+	s           string
+}
```
peterbourgon, line 17: Embedding an io.Reader in this way allows callers to treat Entry as an io.Reader directly, which doesn't seem to be correct, as methods like String consume from that Reader and mutate internal state in a way that callers aren't able to do.

This should probably be something like

```suggestion
type Entry struct {
	src  io.Reader
	read bool
	str  string
}
```

joeshaw: Opened a follow-up issue on this. https://github.com/fastly/compute-sdk-go/issues/40

PR #38, opened 2023-05-04
use SendAsync, not SendAsyncStreaming, for requests without a body

joeshaw:
When the request's ManualFramingMode is false, SendAsyncStreaming streams the request body to the backend using "Transfer-Encoding: chunked".  This is the default behavior we want for requests with a body.  For requests without a body, we want to avoid unnecessary chunked encoding, and have observed servers that error when seeing it in certain contexts.  Calling SendAsync instead causes the entire body to be buffered (in this case, zero bytes) and "Content-Length: 0" to be sent instead.

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -415,7 +415,7 @@ func (req *Request) sendABIRequestAsync(backend string) (*fastly.PendingRequest,
 	// instead causes the entire body to be buffered (in this case, zero
 	// bytes) and "Content-Length: 0" to be sent instead.
 	sendFn := req.abi.req.SendAsyncStreaming
-	if req.abi.body == nil {
+	if req.Body == nil {
 		sendFn = req.abi.req.SendAsync
 	}
 
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -281,7 +281,6 @@ func (req *Request) AddCookie(c *Cookie) {
 // that have been preconfigured in your service, regardless of their URL. Once
 // sent, a request cannot be sent again.
 func (req *Request) Send(ctx context.Context, backend string) (*Response, error) {
-
 	if req.abi.req == nil && req.abi.body == nil {
 		//  abi request not yet constructed
 		if err := req.constructABIRequest(); err != nil {
@@ -289,7 +288,7 @@ func (req *Request) Send(ctx context.Context, backend string) (*Response, error)
 		}
 	}
 
-	abiPending, abiReqBody, err := req.sendABIRequestAsyncStreaming(backend)
+	abiPending, abiReqBody, err := req.sendABIRequestAsync(backend)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


fsthttp/request.go
```diff
@@ -402,12 +401,25 @@ func (req *Request) constructABIRequest() error {
...
 
-func (req *Request) sendABIRequestAsyncStreaming(backend string) (*fastly.PendingRequest, *fastly.HTTPBody, error) {
+func (req *Request) sendABIRequestAsync(backend string) (*fastly.PendingRequest, *fastly.HTTPBody, error) {
 	if req.sent {
 		return nil, nil, fmt.Errorf("request already sent")
 	}
 
-	abiPending, err := req.abi.req.SendAsyncStreaming(req.abi.body, backend)
+	// When SendAsyncStreaming is called, it streams the request body to
+	// the backend using "Transfer-Encoding: chunked".  For requests
```
acfoltzer, line 410: Note that this is only the case with the default framing headers mode; see https://docs.rs/fastly/latest/fastly/http/enum.FramingHeadersMode.html

This is definitely the right change to make, but for applications which need to write to the body after sending the headers _and_ want a `content-length`, the manual framing headers mode is available.

joeshaw: Good point, I'll add this nuance to the comments.  (We support framing headers mode in the SDK.)

dkegel-fastly:
For https://fastly.atlassian.net/browse/ESEC-393

fsthttp/request.go
```diff
@@ -402,12 +401,25 @@ func (req *Request) constructABIRequest() error {
...
+	// chunked".  This is the default behavior we want for requests with
+	// a body.  For requests without a body, we want to avoid
+	// unnecessary chunked encoding, and have observed servers that
+	// error when seeing it in certain contexts.  Calling SendAsync
+	// instead causes the entire body to be buffered (in this case, zero
+	// bytes) and "Content-Length: 0" to be sent instead.
+	sendFn := req.abi.req.SendAsyncStreaming
+	if req.Body == nil {
+		sendFn = req.abi.req.SendAsync
+	}
```
joeshaw, line 420: Many `io.Reader` types in the standard library expose a `Len() int` method when constructed from an in-memory buffer (`strings.Reader, `strings.Builder`, `bytes.Buffer` and `bytes.Reader` are all examples).  One possible future enhancement: do a type cast here and also use `SendAsync` when the actual request body has a known length.

dkegel-fastly (approved):
Worked first try.  Ship it :-)

fsthttp/request.go
acfoltzer: That sounds like a good plan. People tend to be surprised when chunked encoding shows up, even though it's "allowed" and a body is expected (unlike the issue giving rise to this change). We try throughout xqd and the Rust SDK to give the body a fixed length when possible to get `content-length` applied.

joeshaw: https://github.com/fastly/compute-sdk-go/pull/44 implements this

PR #44, opened 2023-05-24
don't stream certain in-memory io.Readers

joeshaw:
`bytes.Buffer`, `bytes.Reader`, and `strings.Reader` are all `io.Reader`
implementations on an in-memory buffer.  Their sizes are known ahead of
time, and these types tend to be used for small payloads.  We can send
these payloads with a `Content-Length` rather than using chunked
encoding by sending them with `SendAsync` rather than
`SendAsyncStreaming`.

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -3,11 +3,13 @@
 package fsthttp
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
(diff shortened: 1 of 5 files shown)


joeshaw:
Marked this as a draft because I'm not 100% convinced we definitely want to do this, but it's the implementation following my comment in https://github.com/fastly/compute-sdk-go/pull/38#discussion_r1200696700.

joeshaw:
One possible other adjustment would be to actually examine the size of the buffers (all three types implement `Len() int`) and still stream if over some size threshold.  I'm not sure how much that matters though.

fsthttp/request.go
```diff
@@ -288,33 +290,54 @@ func (req *Request) Send(ctx context.Context, backend string) (*Response, error)
...
+	// certain contexts.
+	//
+	// For requests where the body is an io.Reader implementer where the
+	// size is known in advance, we want to send that along with
+	// Content-Length as well.  Those types are *bytes.Buffer,
+	// *bytes.Reader, and *strings.Reader.
+	//
+	// For all other requests, we stream with chunked encoding.
+	switch underlyingReaderFrom(req.Body).(type) {
+	case nil, *bytes.Buffer, *bytes.Reader, *strings.Reader:
```
dkegel-fastly, line 319: maybe check whether it implements the Lenner interface (which I just made up)?

joeshaw: I thought about this, and I checked the stdlib for all `io.Reader`s that implemented that interface, and these are the only 3 exported ones that came up.  (There are a handful of internal ones, particular in the `net/http` package.)

Ultimately I decided not to do this because I felt that having `Len() int` wasn't a strong enough contract for an in-memory buffer.  You could imagine an `io.Reader` that streams data from the network, but the size is known ahead of time because it was sent by the client in something akin to `Content-Length` and it was exposed this way.  We'd want to stream the data in that case.

joeshaw:
I decided not to limit it based on length because the Rust SDK doesn't do anything like that currently.  (It implements the `From` trait for several string/byte slice/byte vector types and they also just copy.)  However, if we wanted to add it later, we could add a limit for it, set it to a sensible default (64k?) and only copy if it's within the limit.

joeshaw:
Moving back to draft, as this needs some work.

JakeChampion (approved):

PR #46, opened 2023-05-26
add LimitExceeded FastlyStatus code; print unknown codes

acfoltzer:
LimitExceeded was added last year, and printing unknown codes will help when debugging situations when the definitions get out of sync.

I do not have a Go dev environment set up, so I have not compiled or tested this 😬 but hopefully it's a small enough addition that CI will be happy.

### DIFF
--- a/internal/abi/fastly/types.go
+++ b/internal/abi/fastly/types.go
@@ -113,7 +113,7 @@ func (s FastlyStatus) String() string {
 	case FastlyStatusLimitExceeded:
 		return "LimitExceeded"
 	default:
-		return fmt.Printf("unknown error code: %d", s)
+		return fmt.Sprintf("unknown error code: %d", s)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/abi/fastly/types.go
```diff
@@ -106,8 +110,10 @@ func (s FastlyStatus) String() string {
 		return "HTTPHeadTooLarge"
 	case FastlyStatusHTTPInvalidStatus:
 		return "HTTPInvalidStatus"
+	case FastlyStatusLimitExceeded:
+		return "LimitExceeded"
 	default:
-		return "unknown"
+		return fmt.Sprintf("unknown error code: %d", s)
```
dgryski, line 116: I think something like
```
               return "FastlyStatus(" + strconv.Itoa(int(s)) + ")"
```
is a bit more Go like.  (Although note you'll need to add `strconv` as an import)

acfoltzer:
@joeshaw is going to absorb the relevant bits of this draft into his current branch. Joe, I'm gonna tag you as assignee here, feel free to close this once it's no longer useful for you, and thanks for taking it 🙏🏻

PR #47, opened 2023-05-30
Implement Core Cache and Purge APIs

joeshaw:
The Core Cache API (in the `cache/core` package) exposes the primitive operations required to implement high-performance cache applications with advanced features such as request collapsing, streaming miss, revalidation and surrogate key purging.

While this API contains affordances for some HTTP caching concepts such as Vary headers and stale-while-revalidate, this API is not suitable for HTTP caching out-of-the-box.  Future SDK releases will add a more customizable HTTP Cache API with support for customizable read-through caching, freshness lifetime inference, conditional request evaluation, automatic revalidation, and more.

In the simplest cases, the top-level `Insert` and `Lookup` functions are used for one-off operations on a cached object, and are appropriate when request collapsing and revalidation capabilities are not required. The API also supports more complex uses via `Transaction`, which can collapse concurrent lookups to the same item, including coordinating revalidation.

Example uses of this API are included in `_examples/corecache` and in `cache/core/core_test.go`.

This PR also adds support for surrogate key purging in the `purge` package.

Other minor tweaks:
* Bumps the minimum Go version to 1.18
* Adds the `Abandon` hostcall to `fastly.HTTPBody`, which is important for the core cache API
* Adds the `LimitExceeded` `FastlyStatus` (Closes #46)
* Fixes an off-by-one error in the comment witx for cache_override_tag

### DIFF
--- a/cache/core/core_test.go
+++ b/cache/core/core_test.go
@@ -1,0 +1,156 @@
+package core_test
+
+import (
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/fastly/compute-sdk-go/cache/core"
+)
+
+func ExampleLookup() {
+	// f is a core.Found value, representing a found cache item.
+	// core.ErrNotFound is returned if the item is not cached.
+	f, err := core.Lookup([]byte("my_key"), core.LookupOptions{})
+	if err != nil {
+		panic(err)
+	}
+	defer f.Body.Close()
+
+	// The contents of the cached item are in the Found's Body field.
+	cachedStr, err := io.ReadAll(f.Body)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("The cached value was: %s", cachedStr)
+}
+
+func ExampleInsert() {
+	const (
+		key      = "my_key"
+		contents = "my cached object"
+	)
+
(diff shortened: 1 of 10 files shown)


cache/core/core_test.go
```diff
@@ -0,0 +1,156 @@
...
+	fmt.Printf("The cached value was: %s", cachedStr)
+}
+
+func ExampleInsert() {
+	const (
+		key      = "my_key"
+		contents = "my cached object"
+	)
+
+	// w is a core.WriteCloseAbandoner, a superset of io.CloseWriter.
```
dgryski, line 35: `s/io.CloseWriter/io.WriteCloser/`

dgryski (changes requested):
Typo in a comment, but otherwise this still LGTM.

cache/core/core_test.go
joeshaw: Thanks.  Fixed this in a force-push, 8d4d09f.  https://github.com/fastly/compute-sdk-go/compare/3178ed9e2ef70d880148a7fe0bc4520645ade5a4..8d4d09ffd8cec0b325d43866ffeae113400dce44

dgryski (approved):
LGTM

PR #51, opened 2023-06-06
fsthttp: add RegisterDynamicBackend

dgryski:


fsthttp/backend.go
```diff
@@ -0,0 +1,84 @@
...
+		abiOpts.ConnectTimeout(opts.ConnectTimeout)
+	}
+
+	if opts.FirstByteTimeout != 0 {
+		abiOpts.FirstByteTimeout(opts.FirstByteTimeout)
+	}
+
+	if opts.BetweenBytesTimeout != 0 {
+		abiOpts.BetweenBytesTimeout(opts.BetweenBytesTimeout)
+	}
```
joeshaw, line 52: Just confirming: zero values aren't valid in the hostcalls?

fsthttp/backend.go
```diff
@@ -0,0 +1,84 @@
...
+type TLSVersion uint32
+
+const (
+	TLSVersion1   TLSVersion = 0
+	TLSVersion1_1 TLSVersion = 1
+	TLSVersion1_2 TLSVersion = 2
+	TLSVersion1_3 TLSVersion = 3
+)
+
+type BackendOptions struct {
```
joeshaw, line 18: It looks like we also need a `UseSSL bool` in here.

fsthttp/backend.go
```diff
@@ -0,0 +1,84 @@
...
+	HostOverride        string
+	ConnectTimeout      time.Duration
+	FirstByteTimeout    time.Duration
+	BetweenBytesTimeout time.Duration
+	SSLMinVersion       TLSVersion
+	SSLMaxVersion       TLSVersion
+	CertHostname        string
+	CACert              string
+	Ciphers             string
```
joeshaw, line 27: We should document the expected format for ciphers.  Are these colon-delimited OpenSSL-named ciphers?

(Docs for all of these would be good, but it's a must-have for `Ciphers`)

fsthttp/backend.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"time"
+
+	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
+)
+
+type TLSVersion uint32
+
+const (
+	TLSVersion1   TLSVersion = 0
```
joeshaw, line 12:
```suggestion
	TLSVersion1_0 TLSVersion = 0
```

I recommend being explicit about calling it 1.0, since that seems to be the convention and there are incompatibilities between each minor version.

joeshaw (changes requested):
A few requested changes below but looks good overall.

Not necessary for this PR, but what do you think about creating a proper `Backend` type a la the [Rust SDK](https://docs.rs/fastly/latest/fastly/backend/struct.Backend.html) that includes accessors for these things (as well as healthchecks, etc.)?

fsthttp/backend.go
```diff
@@ -0,0 +1,84 @@
...
+		abiOpts.BetweenBytesTimeout(opts.BetweenBytesTimeout)
+	}
+
+	if opts.SSLMinVersion != 0 {
+		abiOpts.SSLMinVersion(fastly.TLSVersion(opts.SSLMinVersion))
+	}
+
+	if opts.SSLMaxVersion != 0 {
+		abiOpts.SSLMaxVersion(fastly.TLSVersion(opts.SSLMaxVersion))
+	}
```
joeshaw, line 60: Because `TLSVersion1` has the zero value, it makes these a little awkward.

What is the default minimum version if not provided?  If it's not TLSv1.0 then it's not possible to set it to that, because it has the zero value.

And not that someone would want to do this, but it's not possible to set TLSv1.0 as the maximum version either.

dgryski:
My plan for the `Backend` type was when the rest of the dynamic backend bits were added, not just the `Register` call.

fsthttp/backend.go
dgryski: Maybe that's what the `UseSSL` bool is for?  I'll see what the JS and Rust SDKs do.

joeshaw: I think the `UseSSL` bool affects whether these apply at all, but I think it's orthogonal to the issue here, which is that the zero value has meaning separate from omission.

Let's suppose the default minimum TLS version is 1.2.  (A reasonably safe choice!)

By way of example, if the code stays as it is,

```
if opts.SSLMinVersion != 0 {
	abiOpts.SSLMinVersion(fastly.TLSVersion(opts.SSLMinVersion))
}
```

then it is not possible to set the minimum version to TLSv1.0.

If it is changed to be,

```
if opts.UseSSL {
	abiOpts.SSLMinVersion(fastly.TLSVersion(opts.SSLMinVersion))
}
```

then by default it overrides the system's default minimum version, lowering it from 1.2 to 1.0 even though that's probably not what the user wants.

I think we need to either:
1. Change the constants so that 0 is "no change" and `TLSVersion_1_0` is 1 (and the others bumped up as well)
2. Switch to using pointers, so `nil` is the "no change" value.  (Ew.)
3. Switch to using a builder pattern for the API

I think option 1 matches the style of the rest of the SDK best.  Since it's a type and constant value defined in the `fsthttp` package, it's not a big deal (although it's a little gross) to decrement the value by 1 before casting it to `fastly.TLSVersion`.  (An alternative would be to `switch` on a conversion between the types, which is more verbose but less gross.)

joeshaw: And I'm not sure about the JS SDK but the Rust SDK uses a builder pattern for this (and most other APIs in the SDK)

dgryski: Yeah, I wonder if I should switch this to a builder-type API.  That would solve a bunch of these issues.  Let me think a bit more about this.

dgryski:
Switched to a builder API.  Still need to add docs though.

dgryski:
Added docs.  I wonder if we want to change the return value to a `(Backend, error)` (even if the backend type is a string wrapper) just to future proof this for when we *do* add querying backends.

joeshaw:
> I wonder if we want to change the return value to a `(Backend, error)` (even if the backend type is a string wrapper) just to future proof this for when we _do_ add querying backends.

Were you thinking that `Backend` would be a `struct` or just a `string`?  I worry about boxing ourselves in with the latter.

I think returning a `Backend` here is a good idea, since we know we'll want it later.

dgryski:
Wasn't sure if I was going to `type Backend string` or `type Backend struct { b string }`.  In any case we're going to need a `String()` method to pass it along to `Send()`, and yes having it as a struct does give us a bit more wiggle room to cache things going forward.  I'll go ahead and add the `struct` type.

joeshaw (approved):
👍

PR #53, opened 2023-06-19
internal: kv store hostcalls are still "object store"

dgryski:


dgryski:
We didn't catch this in CI because while we only have *examples* using kv/object store (which are only tested to make sure they compile), no integration tests to make sure they actually work.

dennismartensson:
👍 We are working on a new set of kv store hostcalls that are async and hold almost the full set of kv store  functionality.

PR #54, opened 2023-06-20
Improve errors on failure

fgsch:
The error names are longish, happy to go for something shorter.

We should probably add other errors, e.g., https://docs.rs/fastly/0.9.4/src/fastly/backend/builder.rs.html#231, but I'd leave that for another PR.

### DIFF
--- a/fsthttp/backend.go
+++ b/fsthttp/backend.go
@@ -12,8 +12,8 @@ var (
 	// create dynamic backends.
 	ErrDynamicBackendDisallowed = errors.New("dynamic backends not supported for this service")
 
-	// ErrDynamicBackendNameInUse indicates the backend name is already in use.
-	ErrDynamicBackendNameInUse = errors.New("backend name already in use")
+	// ErrBackendNameInUse indicates the backend name is already in use.
+	ErrBackendNameInUse = errors.New("backend name already in use")
 )
 
 type TLSVersion uint32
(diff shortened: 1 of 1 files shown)


Integralist (approved):

fsthttp/backend.go
```diff
@@ -1,11 +1,21 @@
...
 	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
 )
 
+var (
+	// ErrDynamicBackendDisallowed indicates the service is not allowed to
+	// create dynamic backends.
+	ErrDynamicBackendDisallowed = errors.New("dynamic backends not supported for this service")
+
+	// ErrDynamicBackendNameInUse indicates the backend name is already in use.
+	ErrDynamicBackendNameInUse = errors.New("backend name already in use")
```
joeshaw, line 16:
```suggestion
	// ErrBackendNameInUse indicates the backend name is already in use.
	ErrBackendNameInUse = errors.New("backend name already in use")
```

I suggest this because (1) it makes the name shorter and (2) the names have to be unique across all backends, including static ones.

joeshaw (commented):
LGTM, one naming suggestion below.

fsthttp/backend.go
```diff
@@ -111,9 +121,21 @@ func RegisterDynamicBackend(name string, target string, options *BackendOptions)
...
 		abiOpts = &fastly.BackendConfigOptions{}
 	}
+	err := fastly.RegisterDynamicBackend(name, target, abiOpts)
+	if err != nil {
+		status, ok := fastly.IsFastlyError(err)
+		switch {
+		case ok && status == fastly.FastlyStatusUnsupported:
+			return nil, ErrDynamicBackendDisallowed
+		case ok && status == fastly.FastlyStatusError:
+			return nil, ErrDynamicBackendNameInUse
```
joeshaw, line 131:
```suggestion
			return nil, ErrBackendNameInUse
```

PR #56, opened 2023-06-23
fsthttp,internal: add backend query API

dgryski:


fsthttp/backend.go
```diff
@@ -27,6 +27,12 @@ type Backend struct {
 	target string
 }
 
+func NewBackend(name string) *Backend {
+	return &Backend{
+		name: name,
+	}
+}
```
joeshaw, line 34: Is the use case here to instantiate a `Backend` for an existing static backend?  If so, what do you think about calling this something like `BackendForName`, checking `Exists()` in the constructor and returning an `ErrBackendNotFound` if it's not there?

joeshaw: And furthermore, for static backends does it make sense to set `target` to the value of `BackendGetHost`?  I dunno how that's supposed to map, exactly.

dgryski: Good point, that probably makes more sense.

dgryski:
Regarding the API, one thing I was struggling with was if we should actually expose all the different calls, or just have a single end point that gets all the information for a given backend.

joeshaw:
> Regarding the API, one thing I was struggling with was if we should actually expose all the different calls, or just have a single end point that gets all the information for a given backend.

`IsHealthy` is the only thing that looks to be on-demand to me.  The rest of the stuff I think you could populate from the `BackendFromName` call.  (This is what we do for the `Found` in the core cache API.  https://github.com/fastly/compute-sdk-go/blob/main/cache/core/core.go#L122)

dgryski:
PTAL

fsthttp/backend.go
```diff
@@ -47,6 +83,33 @@ func (b *Backend) Target() string {
...
+
+func (b *Backend) Port() (int, error) {
+	return fastly.BackendGetPort(b.name)
+}
+
+func (b *Backend) OverrideHost() (string, error) {
+	return fastly.BackendGetOverrideHost(b.name)
+}
+
```
joeshaw, line 112: Not a big deal, but what do you think about making these fields rather than methods?  I don't think these can be changed after the fact, and it'd eliminate the need to do error checking by users.  They could be populated in `BackendFromName` and `RegisterDynamicBackend` which need to do error handling anyway.

dgryski: That makes sense.  Just realized I didn't fill in all of the possible query fields here.  I think I might try to add Getters to the BackendOptions type and fully populate that on instantiation.

dgryski:
PTAL

joeshaw (approved):
love it 👍

PR #58, opened 2023-07-07
fsthttp: add HandoffWebsocket, HandoffFanout host calls

dgryski:


dgryski:
Need to tag this feature as experimental, possible with a `fastly_experimental` build tag or equivalent.   The non-experimental versions could either not exist (ugly) or return an error (slightly better).

joeshaw (approved):
👍

dgryski:
We should figure out why this test is flaky: https://github.com/fastly/compute-sdk-go/actions/runs/5515994683/jobs/10056895134?pr=58#step:7:537

dgryski:
PTAL.  Now with fine-grained experimental tags.

fsthttp/response.go
```diff
@@ -111,6 +111,26 @@ type ResponseWriter interface {
...
+
+	// HandoffFanout passes the request through the Fanout GRIP proxy and on to
+	// a backend.
+	//
+	// This can only be used on services that have the Fanout feature enabled.
+	//
+	// The sending completes in the background. Once this method has been
+	// called, no other response can be sent to this request, and the
+	// application can exit without affecting the send.
+	HandoffFanout(backend string) error
```
joeshaw, line 133: I'm not necessarily advocating for this but: if these are experimental they could live in a separate build-tagged interface and not be on `fsthttp.ResponseWriter`.  The upside is that if you tried to use them you'd get a build failure rather than a runtime error.  The downside is that you'd need a migration path for when these move from being experimental to being always included.

The general shape of things would be something like:

```go
//go:build fastly_experimental || fastly_experimental_fsthttp || fastly_experimental_fsthttp_handoff

package fsthttp

type ExperimentalResponseWriter interface {
	ResponseWriter
	HandoffWebsocket(backend string) error
	HandoffFanout(backend string) error
}

func Experimental(rw ResponseWriter) ExperimentalResponseWriter {
	return rw.(ExperimentalResponseWriter)
}

func (resp *responseWriter) HandoffWebsocket(backend string) error { ... }
func (resp *responseWriter) HandoffFanout(backend string) error { ... }
```

And calling code that wanted to use it would call `Experimental(rw).HandoffWebsocket(backend)`.

At some point those methods would be moved to the main `ResponseWriter` and users could (but don't have to) drop the `Experimental` wrapper call.

Anyway, again, not necessarily advocating for this but it's an idea that popped up when I first saw the runtime errors.

joeshaw (approved):
LGTM.  I have an alternative idea below but I'm not convinced it's better.  Curious to know what you think.

fsthttp/response.go
dgryski: I'd probably call it a `HandoffResponseWriter`, but yeah.  I was trying to figure out how to have these named-spaced more obviously experimental (like the rust ones are).  This seems like a nice approach.

awilliams-fastly: Could these types (e.g. `HandoffResponseWriter`) be in a separate package? Maybe within the existing [x](https://pkg.go.dev/github.com/fastly/compute-sdk-go@v0.1.5/x) package? You could then potentially drop the build tags for better discoverability & ease of use.

joeshaw: Ooh, I like that idea.  The benefit of a separate package is that you can move the wrapper function (`Experimental()` in my example) into that other package and not pollute the namespace in `fsthttp`.

The main `fsthttp.responseWriter` would still implement the interface, and once it's no longer experimental the methods would be added to the main `ResponseWriter` interface.

dgryski:
PTAL.  I've moved the interface definitions to an experimental package.

dgryski:
lol, so tried to finally test this and it turns out it's a method on a Request, *not* a method on the Response.  So ... uh ... this code is totally wrong and all the design discussions are kind of moot.  Let me see what I can do to fix this ...

dgryski:
PTAL.  Much simpler now.

awilliams-fastly (approved):
👍

joeshaw (approved):
👍

PR #61, opened 2023-07-13
Implement Simple Cache API

joeshaw:
This introduces a new Simple Cache API in the `cache/simple` package.

This provides a simpler interface to inserting and retrieving entries from Fastly's cache: `Get`, `GetOrSet`, and `Purge`.

Cache operations are local to the Fastly POP serving the request, but purges can also be performed globally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Added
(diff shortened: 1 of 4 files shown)


dgryski (approved):
LGTM

PR #63, opened 2023-07-18
rework integration tests to be driven by `tinygo test`

joeshaw:
Integration tests were previously driven by a Javascript-based GitHub
Action.  These ended up being pretty flaky and were not easy to run
locally.  As a result, they were difficult to write and folks waited
until CI to see if they passed.

Viceroy (as of 0.4.0 or so) can act as a test runner by passing in a
.wasm file in a special run-once mode.  Combined with a special TinyGo
target config (compute-at-edge.json) that uses this Viceroy mode as the
"emulator", we can run the integration tests with `tinygo test` like so:

    tinygo test -target=compute-at-edge.json ./integration_tests/...

The tests are build flagged so that they aren't run outside of a WASI
environment, so doing things like `go test ./...` won't attempt to run
them.

The integration tests themselves have been changed so that they contain
not only the handlers but also calling the handlers and checking the
results.  The expected results were previously defined in a JSON file
and checked by the JS runner.

A lot of the churn here is removing the old github action, and moving
the integration tests from `_integration_tests` to `integration_tests`
so that they are picked up by `tinygo test`.

The `rtlog` integration test was the only one that was dropped.  Because
the tests don't have visibility into the Viceroy runtime, it's not
possible to capture the logs.


joeshaw:
New backend integration test is failing because Viceroy is erroneously returning `Inval` when it should return `None` for its internal `ValueAbsent` error.  https://github.com/fastly/Viceroy/pull/291

compute-at-edge.json
```diff
@@ -0,0 +1,26 @@
+{
+	"llvm-target":   "wasm32-unknown-wasi",
```
dgryski, line 2: We can probably use `"inherits":"wasi",` here and just override the emulator line.

dgryski (approved):
This is fantastic.  LGTM.

compute-at-edge.json
joeshaw: nice!  I did not know about this, it makes things so much cleaner!  Did that in 9341fec.

PR #64, opened 2023-07-20
internal: update for Go 1.21rc3

dgryski:
perl rewrite.pl hostcalls_guest.go >guest.tmp && mv guest.tmp hostcalls_guest.go && goimports -w hostcalls_guest.go

### DIFF
--- a/internal/abi/fastly/hostcalls_guest.go
+++ b/internal/abi/fastly/hostcalls_guest.go
@@ -1,5 +1,4 @@
-//go:build tinygo.wasm && wasi && !nofastlyhostcalls
-// +build tinygo.wasm,wasi,!nofastlyhostcalls
+//go:build ((tinygo.wasm && wasi) || wasip1) && !nofastlyhostcalls
 
 // Copyright 2022 Fastly, Inc.
 
@@ -29,14 +28,14 @@ func init() {
 //	   (result $err $fastly_status))
 //	)
 //
-//go:wasm-module fastly_abi
-//export init
+//go:wasmimport fastly_abi init
 //go:noescape
 func fastlyABIInit(abiVersion prim.U64) FastlyStatus
 
 // TODO(pb): this doesn't need to be exported, I don't think?
 // Initialize the Fastly ABI at the given version.
 //func Initialize(version uint64) error {
+
 //	return fastlyABIInit(version).toError()
 //}
 
@@ -65,27 +64,26 @@ func fastlyABIInit(abiVersion prim.U64) FastlyStatus
 //	   (result $err $fastly_status)
 //	 )
 //
-//go:wasm-module fastly_uap
-//export parse
+//go:wasmimport fastly_uap parse
 //go:noescape
 func fastlyUAPParse(
-	userAgent prim.Wstring,
+	userAgentData prim.Pointer[prim.U8], userAgentLen prim.Usize,
 
-	family *prim.Char8,
+	family prim.Pointer[prim.Char8],
 	familyLen prim.Usize,
-	familyNWrittenOut *prim.Usize,
+	familyNWrittenOut prim.Pointer[prim.Usize],
 
-	major *prim.Char8,
+	major prim.Pointer[prim.Char8],
 	majorLen prim.Usize,
-	majorNWrittenOut *prim.Usize,
+	majorNWrittenOut prim.Pointer[prim.Usize],
 
-	minor *prim.Char8,
+	minor prim.Pointer[prim.Char8],
 	minorLen prim.Usize,
-	minorNWrittenOut *prim.Usize,
+	minorNWrittenOut prim.Pointer[prim.Usize],
 
-	patch *prim.Char8,
+	patch prim.Pointer[prim.Char8],
 	patchLen prim.Usize,
-	patchNWrittenOut *prim.Usize,
+	patchNWrittenOut prim.Pointer[prim.Usize],
 ) FastlyStatus
 
 // ParseUserAgent parses the user agent string into its component parts.
@@ -98,24 +96,26 @@ func ParseUserAgent(userAgent string) (family, major, minor, patch string, err e
 		patchBuf  = prim.NewWriteBuffer(cap)
 	)
 
+	userAgentBuffer := prim.NewReadBufferFromString(userAgent).Wstring()
+
 	if err := fastlyUAPParse(
-		prim.NewReadBufferFromString(userAgent).Wstring(),
+		userAgentBuffer.Data, userAgentBuffer.Len,
 
-		familyBuf.Char8Pointer(),
+		prim.ToPointer(familyBuf.Char8Pointer()),
 		familyBuf.Cap(),
-		familyBuf.NPointer(),
+		prim.ToPointer(familyBuf.NPointer()),
 
-		majorBuf.Char8Pointer(),
+		prim.ToPointer(majorBuf.Char8Pointer()),
 		majorBuf.Cap(),
-		majorBuf.NPointer(),
+		prim.ToPointer(majorBuf.NPointer()),
 
-		minorBuf.Char8Pointer(),
+		prim.ToPointer(minorBuf.Char8Pointer()),
 		minorBuf.Cap(),
-		minorBuf.NPointer(),
+		prim.ToPointer(minorBuf.NPointer()),
 
-		patchBuf.Char8Pointer(),
+		prim.ToPointer(patchBuf.Char8Pointer()),
 		patchBuf.Cap(),
-		patchBuf.NPointer(),
+		prim.ToPointer(patchBuf.NPointer()),
 	).toError(); err != nil {
 		return "", "", "", "", err
 	}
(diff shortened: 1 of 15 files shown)


internal/abi/fastly/hostcalls_guest.go
```diff
@@ -65,27 +65,26 @@ func fastlyABIInit(abiVersion prim.U64) FastlyStatus
...
-//go:wasm-module fastly_uap
-//export parse
+//go:wasmimport fastly_uap parse
 //go:noescape
 func fastlyUAPParse(
-	userAgent prim.Wstring,
+	userAgentData uint32, userAgentLen uint32,
 
-	family *prim.Char8,
+	family uint32,
```
joeshaw, line 73: Could we define our own pointer type here? `prim.Pointer` maybe?  I'd feel better about that than having just `uint32` all over the place.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -98,24 +97,26 @@ func ParseUserAgent(userAgent string) (family, major, minor, patch string, err e
...
 	)
 
+	userAgentBuffer := prim.NewReadBufferFromString(userAgent).Wstring()
+
 	if err := fastlyUAPParse(
-		prim.NewReadBufferFromString(userAgent).Wstring(),
+		userAgentBuffer.Data, userAgentBuffer.Len,
 
-		familyBuf.Char8Pointer(),
+		uint32(uintptr(unsafe.Pointer(familyBuf.Char8Pointer()))),
```
joeshaw, line 105: and similarly, a conversion function here that did all the necessary casting

internal/abi/prim/prim.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // Usize is an unsigned integer who's size is based on the system architecture.
-type Usize uint
+type Usize uint32
```
joeshaw, line 11: This type probably doesn't make sense anymore.

joeshaw:
I did a quick rebase of this PR on top of the integration test changes I am making in #63 and things look good there:

```
% GOARCH=wasm GOOS=wasip1 go1.21rc3 test -exec "viceroy run -C fastly.toml" ./integration_tests/...
ok  	github.com/fastly/compute-sdk-go/integration_tests/async_select	4.755s
ok  	github.com/fastly/compute-sdk-go/integration_tests/byte_repeater	9.981s
ok  	github.com/fastly/compute-sdk-go/integration_tests/config_store	4.890s
ok  	github.com/fastly/compute-sdk-go/integration_tests/hello_world	4.711s
ok  	github.com/fastly/compute-sdk-go/integration_tests/kvstore	4.864s
ok  	github.com/fastly/compute-sdk-go/integration_tests/request_downstream	4.898s
ok  	github.com/fastly/compute-sdk-go/integration_tests/request_upstream	4.964s
ok  	github.com/fastly/compute-sdk-go/integration_tests/secret_store	4.839s
ok  	github.com/fastly/compute-sdk-go/integration_tests/status	4.894s
ok  	github.com/fastly/compute-sdk-go/integration_tests/streaming_close	10.084s
```

internal/abi/prim/prim.go
dgryski: My plan more or less was to use this PR for getting the compatibility for Go 1.21, and leave any cleanups of the auto-generated code until later.

internal/abi/fastly/hostcalls_guest.go
dgryski: Nice idea.

internal/abi/fastly/hostcalls_guest.go
dgryski: That makes sense.  I was just trying to get this working, but now that I have the conversion script working I can tweak exactly what we're generating here.

joeshaw:
Doing a little C@E testing I found a couple of things that were missing (unfortunately Viceroy doesn't support the Cache APIs yet so I had to test this against a live service, and the integration tests don't catch it).

In `internal/abi/fastly/types.go` there are a couple of types that have pointers in them.  One example is `cacheWriteOptions`:  https://github.com/fastly/compute-sdk-go/blob/main/internal/abi/fastly/types.go#L551-L563

The `*prim.Char8` there need to be converted into uint32, and the assignments updated.  For example, https://github.com/fastly/compute-sdk-go/blob/main/internal/abi/fastly/hostcalls_guest.go#L2774

Looks like there are a few others in `types.go` that need to be updated: `purgeOptions, `backendConfigOptions` as well as the `multiValueHostcall` function.

internal/abi/prim/prim.go
dgryski: Using this type in the definition of `Wstring` now.

dgryski:
Tweaks to the code generator; we now have a `prim.Pointer` type and a `prim.ToPointer()` function.  Every time we remove a pointer from a decl we add a comment saying what the type *was* for future reference.

dgryski:
Bumped tinygo version and updated the structs to remove the pointers.

dgryski:
Rebased to include the `get_port` fix and solve merge conflict.

internal/abi/prim/prim.go
```diff
@@ -25,10 +25,18 @@ type U32 uint32
 // U64 is an unsigned 64 bit integer.
 type U64 uint64
 
+// Pointer is the type of a pointer
+type Pointer uint32
```
joeshaw, line 29: One thing we could do is define this as

```go
type Pointer[_ any] uint32
```

and then instead of putting the old type in a comment it could be in the definition.

```go
func fastlyCacheGetHits(h cacheHandle, d /* *prim.U64 */ prim.Pointer) FastlyStatus
```

becomes

```go
func fastlyCacheGetHits(h cacheHandle, d prim.Pointer[*prim.U64]) FastlyStatus
```

Obviously it has no effect, but it maintains the annotation.

dgryski: Ooh, pretty.

dgryski:
PTAL.  Pointers now use generics to tag the type.  Rebased to include viceroy testing.

internal/abi/fastly/hostcalls_noguest.go
```diff
@@ -1,5 +1,4 @@
-//go:build (!tinygo.wasm && !wasi) || nofastlyhostcalls
-// +build !tinygo.wasm,!wasi nofastlyhostcalls
+//go:build !((tinygo.wasm && wasi) || wasip1) || nofastlyhostcalls
```
joeshaw, line 1: Could you also make this change to the tests under `integration_tests` as well?

internal/abi/prim/prim.go
```diff
@@ -25,10 +25,18 @@ type U32 uint32
...
+// ToPointer turns an arbitrary pointer into a Pointer
+func ToPointer[T any](ptr *T) Pointer[T] {
+	return Pointer[T](uintptr(unsafe.Pointer(ptr)))
+}
+
 // Wstring is a header for a string.
 type Wstring struct {
-	Data uintptr
-	Len  uint
+	Data Pointer[U8]
```
joeshaw, line 38: This is the only one I saw where the new type was different from the old one. 

Just working through this to make sure I understand: this could have been a `*prim.U8` before (a 32-bit value because it's a pointer), but was for whatever reason a `uintptr`, which is a 32-bit value in WASI because it's a 32-bit architecture?

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -1390,50 +1401,8 @@ func (r *PendingRequest) Wait() (response *HTTPResponse, responseBody *HTTPBody,
...
-	respBody *bodyHandle,
-) FastlyStatus
-
-// PendingRequestSelect blocks until one of the provided pending requests is
-// complete. Returns the completed request, and its associated response and
-// response body. If more than one pending request is complete, returns one of
-// them randomly.
-//
-// TODO(pb): is random correct?
-func PendingRequestSelect(reqs ...*PendingRequest) (index int, done *PendingRequest, response *HTTPResponse, responseBody *HTTPBody, err error) {
```
joeshaw, line 1409: If the intent is to fully drop this function, can you remove it from `hostcalls_noguest.go` too?

joeshaw (commented):
Looks pretty good to me.  A few requests (and a question) below.

I've tested this with a handful of services compiled to both TinyGo and Go 1.21rc3 and haven't seen any more issues with this latest PR.

I think this would probably warrant a 0.2.0 version number when it's released.

internal/abi/prim/prim.go
dgryski: Yes.  `uintptr` is the integer version of a pointer.  TinyGo has a 32-bit wasm port, so `uintptr` was effectively a `uint32`, but Go is currently a weird 32/64-bit hybrid, so a `uintptr` is actually a `uint64`.  However, since `wasmtime` is a 32-bit wasm environment, we need to pass all the pointers as 32-bit things instead of 64-bit things.  Our `prim.Pointer` is effectively a `uint32`, so I decided to turn the `uintptr` into a `prim.Pointer[U8]`.

joeshaw: This was helpful, thank you!

internal/abi/fastly/hostcalls_noguest.go
dgryski: Done

joeshaw:
Not necessarily for this PR but the README will need a big overhaul.

dgryski:
Removed the rewrite script from the git history.

PR #66, opened 2023-07-20
fix a handful of backend bugs

### LINKED ISSUE #65: Backend IsHealthy() API is bool, but should return tri-state
`Backend.IsHealthy()` returns `(bool, error)` but really the hostcall returns `$backend_health` which is an enum:

```
(typename $backend_health
    (enum (@witx tag u32)
        $unknown
        $healthy
        $unhealthy))
```

This means that only "unknown" results in `IsHealth()` returning `false`.  "unhealthy" will report true!

We haven't released this API yet so we should fix this before we do.  IMO we should rename it from `IsHealthy` to `Health` to be more idiomatic with the fact that it doesn't return a bool.

joeshaw:
Replace `backend.IsHealthy()` with `backend.Health()`, because it returns an
enum of 3 states (unknown, healthy, unhealthy) and not a bool.

When populating the backend, the runtime will return `FastlyStatusNone`
if a value isn't set.  This isn't an error, so we should ignore it.
Similarly, the Viceroy environment will return `FastlyStatusUnsupported`
for some unsupported fields and we should ignore them too.

The exported names for getting SSL min/max versions were copypasta'd and
returned the wrong values.

Fixes #65.

### DIFF
--- a/fsthttp/backend.go
+++ b/fsthttp/backend.go
@@ -20,6 +20,15 @@ var (
 	ErrBackendNotFound = errors.New("backend not found")
 )
 
+type BackendHealth uint32
+
+// Constants for dynamic backend health status
+const (
+	BackendHealthUnknown   BackendHealth = 0
+	BackendHealthHealthy   BackendHealth = 1
+	BackendHealthUnhealthy BackendHealth = 2
+)
+
 type TLSVersion uint32
 
 // Constants for dynamic backend TLS configuration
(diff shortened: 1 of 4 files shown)


joeshaw:
I found most of these with a new integration test I'll be adding to #63 once I've rebased it on this branch.

dgryski (approved):
LGTM.  Thanks.

PR #68, opened 2023-07-25
return nicer ErrBackendNotFound when sending to unknown backend

joeshaw:
We created this error for the dynamic backend work, but it applies to
Send() as well.

Because Send wraps the underlying FastlyStatus error, we need to support
unwrapping in IsFastlyError.


joeshaw:
Note that Viceroy currently returns a generic `FastlyStatusError` rather than `FastlyStatusInval`, which C@E returns.  See fastly/Viceroy#292.

dgryski (approved):
LGTM

PR #69, opened 2023-07-25
add an example for using the net/http adapter

joeshaw:


_examples/http-adapter/fastly.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["oss@fastly.com"]
+description = ""
+language = "go"
+manifest_version = 2
+name = "example"
+
+[local_server]
+  [local_server.backends]
+    [local_server.backends.ipv4]
+      url = "https://ipv4.joeshaw.org"
```
dgryski, line 13: We use `httpbin.org` for the other examples.  It has an `/ip` end-point we can use.

dgryski (changes requested):
+1 for more examples.  Is there a way to turn this into a unit test (or add a unit test)?

_examples/http-adapter/fastly.toml
joeshaw: 👍 fixed up in 7af48a8

joeshaw:
> Is there a way to turn this into a unit test (or add a unit test)?

I do have a unit test for the adapter in https://github.com/fastly/compute-sdk-go/blob/main/fsthttp/adapter_test.go.  I could create an integration test which uses this, but as it's entirely an SDK thing (no runtime interaction) I'm not sure it buys us anything.

PR #70, opened 2023-07-26
_examples: add fastly.toml for each example, plus a README.

### LINKED ISSUE #67: Examples should be runnable locally with "fastly compute serve"
Items in the `_examples` directory contain code samples but aren't directly runnable.  We should add `fastly.toml` files to each to allow them to be easily run locally with `fastly compute serve` and reasonably easy to deploy with `fastly compute deploy`.

joeshaw:
This mades it so you can run `fastly compute serve` in each example
directory to test them out locally.

Fixes #67

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+_examples/*/bin
+_examples/*/pkg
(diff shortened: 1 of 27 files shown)


dgryski (approved):
LGTM

PR #71, opened 2023-07-27
Add Append method to ResponseWriter

mzhou:
This mirrors the Response::append_body method in the Rust SDK: https://docs.rs/fastly/0.9.5/fastly/http/struct.Response.html#method.append_body

### DIFF
--- a/fsthttp/response.go
+++ b/fsthttp/response.go
@@ -112,8 +112,8 @@ type ResponseWriter interface {
 	// To have an effect on the response, this must be called before any call to Write() or WriteHeader().
 	SetManualFramingMode(bool)
 
-	// Append a body onto the end of this response.
-	// This operation is performed in amortized constant time, and so should always be preferred to reading an entire body and then writing the same contents to another body.
+	// Append a body onto the end of this response. Will fail if passed anything other than a Response's Body field.
+	// This operation is performed in amortized constant time, and so should always be preferred to directly copying a body with io.Copy.
 	Append(other io.ReadCloser) error
 }
 
--- a/fsthttp/response.go
+++ b/fsthttp/response.go
@@ -111,6 +111,10 @@ type ResponseWriter interface {
 	//
 	// To have an effect on the response, this must be called before any call to Write() or WriteHeader().
 	SetManualFramingMode(bool)
+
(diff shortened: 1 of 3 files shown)


fsthttp/response.go
```diff
@@ -111,6 +111,10 @@ type ResponseWriter interface {
 	//
 	// To have an effect on the response, this must be called before any call to Write() or WriteHeader().
 	SetManualFramingMode(bool)
+
+	// Append a body onto the end of this response.
+	// This operation is performed in amortized constant time, and so should always be preferred to reading an entire body and then writing the same contents to another body.
```
joeshaw, line 116: I think we need to make clear that the reader has to come from a Response's Body field, or make the function work with non-`HTTPBody` types [a la `Request.Send()`](https://github.com/fastly/compute-sdk-go/blob/main/fsthttp/request.go#L317-L333).

I might also make it more explicit that this is preferable to `io.Copy`, which I think most Go programmers will understand.  Something like, "This operation is performed in amortized constant time, and so should always be preferred to directly copying a body with io.Copy".

fsttest/recorder.go
```diff
@@ -48,3 +49,9 @@ func (r *ResponseRecorder) Close() error {
 // SetManualFramingMode is a no-op on ResponseRecorder.  It exists to
 // satisfy the fsthttp.ResponseWriter interface.
 func (r *ResponseRecorder) SetManualFramingMode(v bool) {}
+
+// Append is a no-op on ResponseRecorder.  It exists to
+// satisfy the fsthttp.ResponseWriter interface.
+func (r *ResponseRecorder) Append(other io.ReadCloser) error {
+	return nil
```
joeshaw, line 56: This shouldn't be a no-op.  We'll need to copy the bytes into `r.Body` here.

joeshaw: Lots of the existing integration tests use `io.Copy` -- it'd be good to update at least one of them to use `Append` instead to exercise this.

mzhou:
Thanks for the feedback. Updated.

joeshaw (approved):
👍

dgryski:
Thanks.

PR #72, opened 2023-07-28
secretstore: add SecretFromBytes

joeshaw:
This creates a `secretstore.Secret` instance from a byte slice.  This is
for situations in which an API requires a `Secret` but you cannot (for
whatever reason) use a `Store` to store them.


dgryski (approved):
LGTM

PR #78, opened 2023-08-10
CHANGELOG: update for 0.2

dgryski:


joeshaw (approved):
👍

joeshaw:
Only thing is that semver might require it to be `0.2.0`

dgryski:
Fixed to `0.2.0`

PR #81, opened 2023-09-20
Switch to v2 send hostcalls with better error reporting

### LINKED ISSUE #87: All I get is "poll: Fastly error: Error"
I'm trying the basic example proxy example on my own test backend:  `https://autov.stag.v3.crewlinker.com/check/health` but all I get is `poll: Fastly error: error` without any way to debug. It works fine when running it with  `fastly compute serve`

This is basically developing blind and I have to give up on doing anything with compute@edge for now. Hopefully this can be improved

joeshaw:
This switches the Go SDK to use new versions of hostcalls added to improve error reporting when sending requests to backends.

A new `sendErrorDetail` type is created, which implements `error` and translates the various tag values to a human-readable string.

`FastlyError` now holds an `error` field named `Detail` in addition to a `FastlyStatus`.  When `Detail` is set, the `FastlyError` will return the `Detail` error string instead of the `FastlyStatus` string.

`FastlyStatus` gains a new `toSendError` method which takes a `sendErrorDetail` and constructs a `FastlyError` with the detail appropriately set.

As a result the generic "Error" message when encountering backend errors should now be replaced with a more detailed message.

Fixes #87.

### DIFF
--- a/internal/abi/fastly/types.go
+++ b/internal/abi/fastly/types.go
@@ -128,13 +128,16 @@ func (s FastlyStatus) toError() error {
 	}
 }
 
-func (s FastlyStatus) toErrorDetailed(d sendErrorDetail) error {
-	switch s {
-	case FastlyStatusOK:
+func (s FastlyStatus) toSendError(d sendErrorDetail) error {
+	if s == FastlyStatusOK {
 		return nil
-	default:
-		return FastlyError{Status: s, detail: d}
 	}
+
+	if !d.valid() {
+		return FastlyError{Status: s}
+	}
+
+	return FastlyError{Status: s, Detail: d}
 }
 
 // FastlyError decorates error-class FastlyStatus values and implements the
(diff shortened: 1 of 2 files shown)


internal/abi/fastly/types.go
```diff
@@ -127,17 +128,31 @@ func (s FastlyStatus) toError() error {
...
+}
+
 // FastlyError decorates error-class FastlyStatus values and implements the
 // error interface.
 //
 // Note that TinyGo currently doesn't support errors.As. Callers can use the
 // IsFastlyError helper instead.
 type FastlyError struct {
 	Status FastlyStatus
+	detail sendErrorDetail
```
dgryski, line 147: I wonder if we want to replace this with a more generic `error` variable or type that we can use to wrap different Fastly errors as they become available so this struct doesn't end up with a field for each different type of Fastly error returns.

joeshaw: Yeah, or conversely we might want to expose the error tags (and I guess the TLS Alert codes) to callers to be able to switch on that info

joeshaw:
@dgryski PTAL.

dgryski (approved):
LGTM

PR #84, opened 2023-10-24
remove old obsolete template

joeshaw:
We recommend people use `fastly compute init` to get started with the
SDK.


dgryski (approved):
LGTM

PR #91, opened 2023-11-09
fix Changelog format

joeshaw:
The formatting is important, as we autogenerate the changelog on https://developer.fastly.com/reference/changes/go-sdk/ from these.


fgsch (approved):
👍

PR #92, opened 2023-11-16
add a new `device` package for device detection

joeshaw:
Calling `device.Lookup()` returns a `Device` struct with accessor
methods to get information on the user agent such as name/brand/model as
well as booleans for common device types.

Based on the API in the [Rust SDK](https://docs.rs/fastly/latest/fastly/device_detection/index.html) and our VCL variables.

### DIFF
--- a/device/device_detection.go
+++ b/device/device_detection.go
@@ -1,0 +1,142 @@
+// Package device provides device dection based on the User-Agent
+// header.
+package device
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
+)
+
+var (
+	// ErrDeviceNotFound is returned when the device is not found.
+	ErrDeviceNotFound = errors.New("device not found")
+
+	// ErrUnexpected indicates that an unexpected error occurred.
+	ErrUnexpected = errors.New("unexpected error")
+)
+
+type Device struct {
+	info deviceInfo
+}
+
+type deviceInfo struct {
+	Device struct {
+		Name          string `json:"name"`
+		Brand         string `json:"brand"`
+		Model         string `json:"model"`
+		HWType        string `json:"hwtype"`
+		IsEReader     bool   `json:"is_ereader"`
+		IsGameConsole bool   `json:"is_gameconsole"`
+		IsMediaPlayer bool   `json:"is_mediaplayer"`
+		IsMobile      bool   `json:"is_mobile"`
+		IsSmartTV     bool   `json:"is_smarttv"`
+		IsTablet      bool   `json:"is_tablet"`
+		IsTVPlayer    bool   `json:"is_tvplayer"`
+		IsDesktop     bool   `json:"is_desktop"`
+		IsTouchscreen bool   `json:"is_touchscreen"`
+	} `json:"device"`
+}
+
+func Lookup(userAgent string) (Device, error) {
+	var d Device
+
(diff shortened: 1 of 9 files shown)


joeshaw:
This new API requires JSON decoding, but I've tested it on Go and TinyGo, and the integration tests will ensure it works on our supported configurations.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -3387,3 +3387,55 @@ func PurgeSurrogateKey(surrogateKey string, opts PurgeOptions) error {
...
+//go:noescape
+func fastlyDeviceDetectionLookup(
+	userAgentData prim.Pointer[prim.U8], userAgentLen prim.Usize,
+	buf prim.Pointer[prim.Char8],
+	bufLen prim.Usize,
+	nWritten prim.Pointer[prim.Usize],
+) FastlyStatus
+
+func DeviceLookup(userAgent string) ([]byte, error) {
+	buf := prim.NewWriteBuffer(defaultBufferLen)
```
fgsch, line 3414: Is the default buffer length the same in other SDKs?

integration_tests/device_detection/fastly.toml
```diff
@@ -0,0 +1,25 @@
...
+
+
+[local_server]
+[local_server.device_detection]
+format = "inline-toml"
+
+[local_server.device_detection.user_agents]
+[local_server.device_detection.user_agents."Mozilla/5.0 (X11; Linux x86_64; rv:10.0) Gecko/20100101 Firefox/10.0 [FBAN/FBIOS;FBAV/8.0.0.28.18;FBBV/1665515;FBDV/iPhone4,1;FBMD/iPhone;FBSN/iPhone OS;FBSV/7.0.4;FBSS/2; FBCR/Telekom.de;FBID/phone;FBLC/de_DE;FBOP/5]"]
+user_agent = {}
+os = {}
```
fgsch, line 19: Does it make sense to test user_agent and os too?

internal/abi/fastly/hostcalls_guest.go
joeshaw: In our SDK it's 16k.  In Rust it's only 1k.  The Rust SDK is better about using "adaptive buffers", in which you make a hostcall with a certain sized buffer and if it's too small the SDK tells you the size it needs, and you allocate a new buffer to fit it.  We do it for a few hostcalls in the Go SDK (and any new one I write, including this one) but it isn't pervasive.

I can resize this one down to 1k to match it.

fgsch: Thank you for the explanation. I think it's fine. I was more concerned about being smaller, in which cases we would need to do 2 calls. I think we can leave it as is.

integration_tests/device_detection/fastly.toml
fgsch: Not a blocker, btw.

fgsch (approved):
Thank you!

device/device_detection.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	if err := json.Unmarshal(raw, &d.info); err != nil {
+		return d, err
+	}
+
+	return d, nil
+}
+
+// Name returns the name of the client device.
```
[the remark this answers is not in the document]
joeshaw: The device stuff is the only stuff we're committing to at the moment.  I force-pushed a commit that removes the user agent and OS stuff for now.

fgsch (approved):
Still 👍

PR #93, opened 2023-11-16
move geolocation parser over to encoding/json

joeshaw:
Now that `encoding/json` is supported on all our supported versions of
TinyGo, we don't need our bespoke JSON parser anymore.

This also simplifies the geolocation integration test: we don't need to
wrap the test in an HTTP request because geo lookups don't require a
request context, just an IP address.


fgsch (approved):
👍  RIP old code 💀

PR #94, opened 2023-11-16
Simplify some integration tests

joeshaw:
Many integration tests don't need to be run inside a request, so we can simplify those by testing them directly.

Also, this renames the `compute-at-edge.json` file to `fastly-compute.json`.  This file is used as a target configuration for running against Viceroy when testing with TinyGo tests.


dgryski (approved):
LGTM

PR #96, opened 2023-12-08
refactor some request code to make the flow simpler & reduce side effects

joeshaw:
I started doing this to bring #44 up-to-date to decide if we really want to do it.  I found the inconsistency on whether to use the struct members confusing, so I removed them.  Another approach would be to go the other way and lean into tracking that state on the Request struct.


JakeChampion:
@joeshaw is this waiting for a review?

joeshaw:
@JakeChampion Yes

PR #97, opened 2023-12-14
cw/geo errors

cee-dub:
geo: Unpacks internal errors using the IsFastlyError helper.

- go.mod: use go 1.20 for comparable constraint
- geo: use IsFastlyError helper to improve error messages

### DIFF
--- a/geo/geodata.go
+++ b/geo/geodata.go
@@ -12,9 +12,6 @@ import (
 )
 
 var (
-	// ErrNotFound indicates there was no geo data returned.
-	ErrNotFound = errors.New("geo: no data found")
-
 	// ErrInvalidIP indicates the input IP was invalid.
 	ErrInvalidIP = errors.New("geo: invalid IP")
 
@@ -51,7 +48,12 @@ func Lookup(ip net.IP) (*Geo, error) {
 		status, ok := fastly.IsFastlyError(err)
 		switch {
 		case ok && status == fastly.FastlyStatusNone:
-			return nil, ErrNotFound
+			// Viceroy <= 0.9.3 returns fastly.FastlyStatusNone when no geolocation
+			// data is available. The Compute production environment instead returns
+			// empty data, which is handled by falling through to code below this switch.
+
+			// TODO: potential breaking change if bumping major version
+			// return nil, ErrNotFound
 		case ok && status == fastly.FastlyStatusInval:
 			return nil, ErrInvalidIP
 		case ok:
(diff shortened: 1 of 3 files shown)


dgryski (approved):
LGTM

geo/geodata.go
```diff
@@ -35,7 +48,17 @@ type Geo struct {
 func Lookup(ip net.IP) (*Geo, error) {
 	buf, err := fastly.GeoLookup(ip)
 	if err != nil {
-		return nil, err
+		status, ok := fastly.IsFastlyError(err)
+		switch {
+		case ok && status == fastly.FastlyStatusNone:
+			return nil, ErrNotFound
```
joeshaw, line 54: Viceroy will return this ([reference](https://github.com/fastly/Viceroy/blob/8f2ea50bff09edead16aeb16b14d6cb731bdf9b7/lib/src/wiggle_abi/geo_impl.rs#L30)) but the Compute platform never does.  We should probably discuss internally how best to reconcile this.

The Compute platform returns an empty buffer, which is handled down below.

Discrepancy aside, I think that introducing `ErrNotFound` would be a breaking change.  Folks might have written code like:

```go
g, err := geo.Lookup(remoteAddr)
if err != nil {
	panic("oh no!")
}
if g != nil {
	// do something with the geo data
}
```

If we introduce an error in the not found case, things will break.

So in my view, we should:
* Change Viceroy to match Compute's behavior (return `nil, nil` in the not-found case)
* Update this patch to drop `ErrNotFound`
* Improve the docs here to explain the behavior

I don't love this, and it's different from what we do elsewhere in the SDK, but the alternative is revving the major version number and this change alone doesn't justify it, IMO.

joeshaw (changes requested):

geo/geodata.go
joeshaw: Err, I should be more clear.  Just introducing `ErrNotFound` is not a breaking change.  It is a little misleading, since it'd only ever be used when run under Viceroy.  Changing the SDK futher to return it when `len(buf) == 0` would be a breaking change, as would changing the compute platform to return it if there's no match.

dgryski: I agree that we should probably have Viceroy match our production environment as close as possible to avoid misleading users.

cee-dub: As a user, I was a bit misled by Viceroy, so I'm in favor of updating it to match production. That could also potentially break people's builds, but at least that would be in dev and not production.

cee-dub:
@joeshaw PTAL, I think I've addressed the feedback.

joeshaw (approved):
👍

PR #99, opened 2024-01-04
don't defer closing the response writer

### LINKED ISSUE #98: panic returns a 200 OK rather than 500 Internal Server Error
With the example program:

```go

package main

import (
	"context"
	"fmt"
	"os"

	"github.com/fastly/compute-sdk-go/fsthttp"
)

func main() {
	fsthttp.ServeFunc(func(ctx context.Context, w fsthttp.ResponseWriter, r *fsthttp.Request) {
		switch r.URL.Path {
		case "/panic":
			panic("panicking quickly")
		case "/exit":
			os.Exit(1)
		}
	})
}
```

Hitting `/panic` returns a `200 OK` when it should return a `500 Internal Server Error`.  Hitting `/exit` does return a 500.

joeshaw:
Closing the response writer signifies to the Compute runtime that the
response has been successfully completed, so we shouldn't defer it in
case the handler panics.

In TinyGo, defers aren't run after panic so this wasn't executed.  In
Go, defers are run.  The result of this was that if a handler panicked
before sending a response, it would send an empty 200 OK response in Go
but a 500 Internal Server Error in TinyGo.

Fixes #98.

### DIFF
--- a/fsthttp/handle.go
+++ b/fsthttp/handle.go
@@ -35,8 +35,8 @@ func Serve(h Handler) {
 		}
 	})
 
-	defer clientResponseWriter.Close()
 	h.ServeHTTP(ctx, clientResponseWriter, clientRequest)
+	clientResponseWriter.Close()
 }
 
 // ServeFunc is sugar for Serve(HandlerFunc(f)).


Integralist (approved):

Integralist:
@dgryski @joeshaw do we know when this will make it's way into a release?

cc @cee-dub

joeshaw:
v1.2.1 was released today.  https://github.com/fastly/compute-sdk-go/releases/tag/v1.2.1

PR #102, opened 2024-02-15
Edge Rater Limiter APIs

joeshaw:
This introduces Edge Rate Limiting functionality in a new `erl` package.

The RateCounter type can be used to increment an event counter and to examine the rate of events per second within a POP over 1, 10, and 60 second windows.  It can also estimate the number of events seen over the past minute within a POP in 10 second buckets.

The PenaltyBox type can be used to track entries that should be penalized for a certain amount of time.

The RateLimiter type combines a rate limiter and penalty box to determine whether a given entry should be penalized based on whether it exceeds a maximum threshold of events per second over a given rate window, defined in a Policy.  Most users can simply use the RateLimiter's CheckRate method rather than needing to interact with RateCounter and PenaltyBox directly.


erl/erl_test.go
```diff
@@ -0,0 +1,104 @@
...
+			// It's probably better to fail open.  Consider logging the
+			// error but continuing to handle the request.
+		} else if block {
+			// The rate limit has been exceeded.  Return a 429 Too Many
+			// Requests response.
+			w.WriteHeader(fsthttp.StatusTooManyRequests)
+			return
+		}
+
+		// Otherwise, continue processing the request.
```
dgryski, line 41: Worth trying to add an example of querying the request rate/count?

dgryski (approved):
LGTM

joeshaw:
I'm going to remove the JSON unmarshaling.  I'm not comfortable committing to that format in the API without more testing/preview use.

erl/erl_test.go
joeshaw: Added in 80f79d2.

joeshaw:
@dgryski PTAL

erl/erl.go
```diff
@@ -0,0 +1,257 @@
...
+type PenaltyBox struct {
+	name string
+}
+
+// OpenPenaltyBox opens a penalty box with the given name, creating it
+// if it doesn't already exist.  The penalty box name may be up to 64
+// characters long.  Entry names in this penalty box are also limited to
+// 64 characters.
+func OpenPenaltyBox(name string) *PenaltyBox {
+	return &PenaltyBox{name: name}
```
cee-dub, line 166: Worth adding a check like this?
```suggestion
	if len(name) > 64 {
		panic("penalty box: name too long")
	}
	return &PenaltyBox{name: name}
```

erl/erl.go
```diff
@@ -0,0 +1,257 @@
...
+type RateCounter struct {
+	name string
+}
+
+// OpenRateCounter opens a rate counter with the given name, creating it
+// if it doesn't already exist.  The rate counter name may be up to 64
+// characters long.  Entry names in this counter are also limited to 64
+// characters.
+func OpenRateCounter(name string) *RateCounter {
+	return &RateCounter{name: name}
```
cee-dub, line 125: Same name length check would apply here as to `OpenPenaltyBox`.

erl/erl.go
```diff
@@ -0,0 +1,257 @@
...
+// not a continuous one.  See [CounterDuration] for more details.
+func (rc *RateCounter) LookupCount(entry string, duration CounterDuration) (uint32, error) {
+	v, err := fastly.RateCounterLookupCount(rc.name, entry, duration)
+	if err != nil {
+		return 0, mapFastlyError(err)
+	}
+	return v, nil
+}
+
+// PenaltyBox is a type that allows entries to penalized for a given
```
cee-dub, line 155:
```suggestion
// PenaltyBox is a type that allows entries to be penalized for a given
```

erl/erl.go
```diff
@@ -0,0 +1,257 @@
...
+// if it doesn't already exist.  The rate counter name may be up to 64
+// characters long.  Entry names in this counter are also limited to 64
+// characters.
+func OpenRateCounter(name string) *RateCounter {
+	return &RateCounter{name: name}
+}
+
+// Increment increments the rate counter for this entry by the given
+// delta value.  The minimum value is 0 and the maximum is 1000.
+func (rc *RateCounter) Increment(entry string, delta uint32) error {
```
cee-dub, line 130: Should we also check these values in Go? It is fairly common in Go to allow "incrementing" by a negative number as well.

erl/erl.go
```diff
@@ -0,0 +1,257 @@
...
+func OpenPenaltyBox(name string) *PenaltyBox {
+	return &PenaltyBox{name: name}
+}
+
+// Add adds an entry to the penalty box for the given time-to-live (TTL)
+// duration.  The minimum value is 1 minute and the maximum is 60
+// minutes.  If an entry is already in the penalty box, its TTL is
+// replaced with the new value.  Entries are automatically evicted from
+// the penalty box when the TTL expires.
+func (pb *PenaltyBox) Add(entry string, ttl time.Duration) error {
```
cee-dub, line 174: Another ask about ergonomics: this time about checking the duration's validity.

cee-dub (approved):
LGTM. Questions are only around Go ergonomics. One typo suggestion. 🥂

joeshaw (commented):
Regarding the validation checks, this is something @mhp and I had a recent conversation about, and the desire was not to push those out to the SDKs to give us the flexibility to make changes later and not need to update them in all places.  (Yes, we would have to change the documentation, but that's a softer change to make.)

More broadly, this is generally the approach the different SDKs take: let the platform inform the SDKs via return values in the hostcalls about invalid input, and this works pretty well (though is not richly informative as to precisely why the value is invalid) and lets us centralize the handling.

dgryski (approved):
LGTM

erl/erl.go
cee-dub: > that's why it's a `uint32` here

Types, how do they work? 😵‍💫

PR #105, opened 2024-03-07
add Request.CloneWithBody and Request.SetBody

### LINKED ISSUE #104: Add Request.CloneWithBody(io.Reader)
SDK version 1.2.1 allowed us to send `Content-Length` instead of `Transfer-Encoding: chunked` for in-memory `io.Reader` implementations `bytes.Buffer`, `bytes.Reader` and `strings.Reader` as the `Request.Body` field.  However, `Request.Body` is actually an `io.ReadCloser` and it is common to wrap the reader in an `io.NopReader` when assigning to the field directly.  This hides the underlying reader from us, and we can't use `Content-Length` for them.

The `NewRequest` function wraps an `io.Reader` in our own `nopCloser` implementation that _does_ allow us to get to the underlying reader. 

One case that came up recently was someone taking the downstream request, reading in the body into a `bytes.Buffer`, and then wanting to send it again to an upstream server.  They did this reasonably with `r.Body = io.NopCloser(bytes.NewBuffer(data))` but this prevents us from sending `Content-Length`.  The workaround for this is messy: constructing a new `Request` and then manually copying all the settings over.  We have `Request.Clone()` but that doesn't include any `Body`, forcing the user to again use `io.NopCloser(bytes.NewBuffer(data))`.

We should make this a little nicer by adding a new `Request.CloneWithBody(io.Reader)` method that will clone a request and wrap the `io.Reader` in our own `nopReader` implementation.

We can never fully solve the problem of assigning to `Body` directly, but perhaps we should expose our `nopReader` type so end users could use it?

joeshaw:
These methods wrap provided `io.Reader`s that aren't already also closers in our own `nopCloser` type, which we already do in NewRequest.  This extends our ability to send `Content-Length` with certain in-memory types to other uses of requests.

We note in the documentation now that using `SetBody` is preferred over setting the `Body` field directly.)

Fixes #104

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -57,12 +57,9 @@ type Request struct {
 	// reads may return immediately with EOF. For outgoing requests, the body is
 	// optional. A body may only be read once.
 	//
-	// It is possible to assign the unread body of the incoming client request
-	// to the body field of a different request. When that second request is
-	// sent, the body will be efficiently streamed from the incoming request.
-	//
-	// It is also possible to assign the unread body of a received response to
-	// the body field of a request, with the same results.
+	// Prefer using the SetBody method over assigning to this value directly,
+	// as it enables optimizations when sending outgoing requests.  See the
+	// SetBody documentation for more information.
 	Body io.ReadCloser
 
 	// Host is the hostname parsed from the incoming request URL.
@@ -116,11 +113,16 @@ func NewRequest(method string, uri string, body io.Reader) (*Request, error) {
 		return nil, err
 	}
 
+	rc, ok := body.(io.ReadCloser)
+	if !ok && body != nil {
+		rc = nopCloser{body}
+	}
+
 	return &Request{
 		Method: method,
 		URL:    u,
 		Header: NewHeader(),
-		Body:   makeBodyFor(body),
+		Body:   rc,
 		Host:   u.Host,
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


joeshaw:
I wonder if we should also have a `SetBody` method and prefer that to assigning to `.Body` directly for the same reason?

cee-dub (commented):
I just made the change in my code to use Clone yesterday, and almost forgot the post body. This is timely!

cee-dub:
I prefer your suggestion of exposing the type, with a clear doc string explaining how it enables the `content-length` header. That leaves the author with the most flexibility.

joeshaw:
I pushed a replacement commit that adds `SetBody`.  You could make the argument that we no longer need `CloneWithBody` now.

fsthttp/request.go
```diff
@@ -209,8 +211,30 @@ func newClientRequest() (*Request, error) {
...
+// If an unread body from an incoming client request is set on an
+// outgoing upstream request, the body will be efficiently streamed from
+// the incoming request.  It is also possible to set the unread body of
+// a received response to the body of a request, with the same results.
+//
+// If the body is set from an in-memory reader such as [bytes.Buffer],
+// [bytes.Reader], or [strings.Reader], the runtime will send the
+// request with a Content-Length header instead of Transfer-Encoding:
+// chunked.
+func (req *Request) SetBody(body io.Reader) {
```
cee-dub, line 227: A bit of a learning question here, but is it possible to have the case of and incoming request with a content-length brought to this body assignment as well?

fsthttp/request.go
```diff
@@ -230,6 +254,14 @@ func (req *Request) Clone() *Request {
 	}
 }
 
+// CloneWithBody returns a copy of the request, with the Body field set
+// to the provided io.Reader.  Its URL will have a nil User field.
+func (req *Request) CloneWithBody(body io.Reader) *Request {
```
cee-dub, line 259: I agree this could be removed now, perhaps by including the two lines of code into the doc of `SetBody`.

fsthttp/request.go
cee-dub: Realized this isn't super clear, but I think `Clone` already handles this by copying the header if present.

joeshaw: Yes, but the runtime will remove `Content-Length` and `Transfer-Encoding` before sending the request unless `ManualFramingMode` is true.

PR #110, opened 2024-04-25
Implement KVStore delete method

athomason:
Note that the synchronous `delete` hostcall doesn't exist unlike `insert` and `lookup`, so this uses `delete_async` and `pending_delete_wait` like the Rust and JS SDKs do.


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.3.1 (2024-04-25)
```
joeshaw, line 1: This should be `## Unreleased` until we tag it, but if we do a release immediately after this lands that's fine.

joeshaw (approved):
👍 LGTM with one note below.

CHANGELOG.md
athomason: Cool, that's my intent.

PR #111, opened 2024-04-25
Switch request reflector backend

athomason:
Requests to http-me.glitch.me occasionally fail with 500 errors, leading to test flakes [1]. @BrooksCunningham suggests using the nearly compatible http.edgecompute.app [2] instead, which seems to be more reliable and substantially faster thanks to being hosted directly on Fastly Compute.

One required change is not sending invalid UTF-8 in the request body, which results in a 500 Internal Server Error.

1. https://github.com/fastly/compute-sdk-go/actions/runs/8839087224/job/24271648933
2. https://github.com/BrooksCunningham/http-me-rust

### DIFF
--- a/integration_tests/request_upstream/fastly.toml
+++ b/integration_tests/request_upstream/fastly.toml
@@ -22,4 +22,4 @@ service_id = ""
       url = "https://example.org/"
 
     [local_server.backends.httpme]
-      url = "https://http-me.glitch.me"
+      url = "https://http.edgecompute.app/anything/"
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM.  What about all the other occurrences of `http-me.glitch.me` in the repo?

joeshaw:
I bulk-changed `httpbin.com` at one point to `http-me.glitch.me`, so a similar bulk change to `http.edgecompute.app` would be a good idea.

athomason:
It would be easy to change the examples with the exception of e.g. `"https://http-me.glitch.me/drip=2?wait=3000", // delay 3s + stream body 2s = 5s`; the delay functionality isn't in http.edgecompute.app. I'll check how much work that would be to add.

dgryski:
Moved to `http-me.fastly.dev` which is a compatible implementation running on Compute.  https://github.com/fastly/compute-sdk-go/pull/171 .

PR #112, opened 2024-06-07
Add a `Has` method to Dictionary

elliottt:
Add a `Has` method to the `Dictionary` struct, which will avoid allocations when it's only desired to check if the value exists.

`Has` uses the same host call as `Get`, but gives it an empty buffer. This way it's able to inspect `FastlyStatus` to determine if the value was present, instead of actually copying the value over.

### DIFF
--- a/internal/abi/fastly/hostcalls_guest.go
+++ b/internal/abi/fastly/hostcalls_guest.go
@@ -2443,7 +2443,7 @@ func (d *Dictionary) Get(key string) (string, error) {
 	return buf.ToString(), nil
 }
 
-// Check if a value exists.
+// Has returns whether a value exists.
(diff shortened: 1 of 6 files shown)


internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2443,6 +2443,32 @@ func (d *Dictionary) Get(key string) (string, error) {
 	return buf.ToString(), nil
 }
 
+// Check if a value exists.
```
dgryski, line 2446: Comment should be `Has returns whether a value exists`

dgryski (changes requested):
Comment tweak and this should have a test added to `./integration_tests/config_store`.

dgryski (approved):
LGTM

PR #113, opened 2024-06-11
integration_tests/config_store: Fix build errors

lpereira:
This fixes the "no new variables on left side of :=" while running the config_store integration tests.

### DIFF
--- a/integration_tests/config_store/main_test.go
+++ b/integration_tests/config_store/main_test.go
@@ -25,7 +25,7 @@ func TestConfigStore(t *testing.T) {
 		t.Errorf("Has reported `true` for a missing key")
 	}
 
-	present, err := d.Has("empty-value")
+	present, err = d.Has("empty-value")
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Looks good to me! Do you know why this didn't fail before?

cee-dub (approved):

cee-dub:
Likely the commands are exiting 0 even if building fails. I have a branch to investigate that I haven't pushed up yet.

joeshaw:
The code was introduced in #112, and the integration tests weren't run as part of CI.  I suspect this is because it was a fork of the repo.

joeshaw:
Integration tests and build examples have this at the top of their GHA yaml:

```
on: [push]
```

whereas validate (which does run) has:

```
on:
  push:
    branches:
      - main
  pull_request:
```

So we probably should add `pull_request` to the other two, but I haven't done any more digging than this.

cee-dub:
Yep, I got that far and discovered the same (discussed in Slack). Thanks for confirming.

I'm doing more cleanup/efficiency work on the actions, but if smaller PRs are preferred I can do incremental changes.

PR #114, opened 2024-06-11
configstore: Implement GetBytes() method

lpereira:
Similar to the Get() method, this one will return the byte slice as is from the Dictionary, without converting it to a string.  May be useful in cases where configuration is stored as binary blobs.

### DIFF
--- a/configstore/configstore.go
+++ b/configstore/configstore.go
@@ -86,26 +86,36 @@ func (s *Store) Has(key string) (bool, error) {
 	return v, nil
 }
 
-// Get returns the item in the config store with the given key.
-func (s *Store) Get(key string) (string, error) {
+// Get returns the item in the config store with the given key, as a byte slice.
+func (s *Store) GetBytes(key string) ([]byte, error) {
 	if s == nil {
-		return "", ErrKeyNotFound
+		return nil, ErrKeyNotFound
 	}
 
-	v, err := s.abiDict.Get(key)
+	v, err := s.abiDict.GetBytes(key)
 	if err != nil {
 		status, ok := fastly.IsFastlyError(err)
 		switch {
 		case ok && status == fastly.FastlyStatusBadf:
-			return "", ErrStoreNotFound
+			return nil, ErrStoreNotFound
 		case ok && status == fastly.FastlyStatusNone:
-			return "", ErrKeyNotFound
+			return nil, ErrKeyNotFound
 		case ok:
-			return "", fmt.Errorf("%w (%s)", ErrUnexpected, status)
+			return nil, fmt.Errorf("%w (%s)", ErrUnexpected, status)
 		default:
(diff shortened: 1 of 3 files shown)


configstore/configstore.go
```diff
@@ -86,26 +86,36 @@ func (s *Store) Has(key string) (bool, error) {
...
 }
+
+// Get returns the item in the config store with the given key.
+func (s *Store) Get(key string) (string, error) {
+	buf, err := s.GetBytes(key)
+	if err != nil {
+		return "", err
+	}
+
+	return string(buf), nil
```
ydnar, line 120: Does this optimize away to the non-allocating byte slice to string conversion?

internal/abi/fastly/hostcalls_noguest.go
```diff
@@ -297,6 +297,10 @@ func OpenDictionary(name string) (*Dictionary, error) {
 	return nil, fmt.Errorf("not implemented")
 }
 
+func (d *Dictionary) GetBytes(key string) ([]byte, error) {
+	return []byte{}, fmt.Errorf("not implemented")
```
ydnar, line 301: Why not `return nil, ...`?

Second, `fmt.Errorf("const string")` should be replaced with a declared `errors.New("...")` in a global or a custom `type errorString string` and a const.

configstore/configstore.go
lpereira: I have patches to TinyGo that will do that, yes.

internal/abi/fastly/hostcalls_noguest.go
cee-dub: This is all over the stubbed implementations without `wasip1` build tags.

lpereira: Sure, I can return `nil`.  (The rest here I just copied what was already there. I didn't want to make changes that weren't related to the change in this PR.)

lpereira: (Patch to return `nil` pushed.)

configstore/configstore.go
cee-dub: I added this to the integration test, and see the following output:
`~/go/src/github.com/fastly/compute-sdk-go/integration_tests/config_store/main_test.go:47:23: object allocated on the heap: escapes at line 52`
``` go
	tb, err := d.GetBytes("twitter")
	if err != nil {
		t.Fatal(err)
	}
	if got, want := tb, "https://twitter.com/fastly"; !bytes.Equal(tb, []byte(want)) {
		t.Errorf("Body = %q, want %q", got, want) // line 52
	}
```

cee-dub: Can push up that commit to test the new func

cee-dub:
See #115

PR #116, opened 2024-06-12
GHA: fix up actions, including integration_tests for PRs

cee-dub:
Actions now runs tests, integration tests, and builds examples on pushes and PRs.

Some of the Go/TinyGo separate actions were consolidated to do both to save time on setup/teardown.

Versions of included actions were bumped.

### DIFF
--- a/.github/workflows/build-examples.yml
+++ b/.github/workflows/build-examples.yml
@@ -1,5 +1,10 @@
 name: Build Examples
-on: [push]
+on:
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-examples.yml
```diff
@@ -1,7 +1,8 @@
 name: Build Examples
 on: [push]
```
joeshaw, line 2: Can you have this run on pull requests as well?

PR #117, opened 2024-06-13
Use known and adaptive buffer sizes to lower generated garbage

### LINKED ISSUE #101: Use adaptive buffers everywhere
The Compute runtime has a concept called "adaptive buffers."  The idea is that when you pass a buffer into a hostcall that's too small for the result, the `buflen` error status is returned, along with the size the buffer would have needed to be to fit the data.  The SDK then allocates an appropriately sized buffer and calls the hostcall again.

Here's an example from the device detection hostcall: https://github.com/fastly/compute-sdk-go/blob/579dadcdcd295da646e6836c5d922eaaa26b89b6/internal/abi/fastly/hostcalls_guest.go#L3414-L3435

We use this pattern for all new development, but much of the older parts of the SDK don't support adaptive buffers.  This pushes the burden onto end users to use the limits API to empirically increase the size of the default buffers to avoid errors.  See #7 for the introduction of the limits API for the impetus for this.

We should audit and fix these calls to use adaptive buffers, which will allow SDK users to automatically get data up to the largest size permitted by our platform without additional work.  Then the limits API could be deprecated.

cee-dub:
Closes #101

To enable merging this branch, we verified which of the following hostcalls **fully** support adaptive buffers (return a `buflen` error with a usable value) in both ExecuteD and Viceroy:

ExecuteD:
- [x] fastly_http_req method_get
- [x] fastly_http_req uri_get
- [x] fastly_dictionary get (partial!) _Full support needs new hostcall that returns the needed size_
- [x] fastly_geo lookup
- [x] fastly_secret_store plaintext
- [x] fastly_device_detection lookup
- [x] fastly_cache get_user_metadata
- [x] fastly_http_req downstream_tls_client_hello

Viceroy:
- [x] fastly_device_detection lookup
- [x] fastly_dictionary get (partial!) _Full support needs new hostcall that returns the needed size_
- [x] fastly_http_req method_get
- [x] fastly_http_req uri_get
- [x] fastly_geo lookup
- [x] fastly_secret_store plaintext
- [ ] fastly_cache get_user_metadata (❌ not implemented in viceroy)
- [ ] fastly_http_req downstream_tls_client_hello (❌ not implemented in viceroy)

Any hostcalls found not to match behavior between ExecuteD and Viceroy will be corrected in Viceroy.

Any hostcalls found not to return a useable value in ExecuteD will be held back in a separate branch until a new version is fully rolled to the fleet.

### DIFF
--- a/internal/abi/fastly/hostcalls_guest.go
+++ b/internal/abi/fastly/hostcalls_guest.go
@@ -638,23 +638,23 @@ func fastlyHTTPReqDownstreamTLSClientHello(
 // DownstreamTLSClientHello returns the ClientHello message sent by the client
 // in the singleton downstream request, if any.
 func DownstreamTLSClientHello() ([]byte, error) {
-	n := DefaultLargeBufLen
-alloc:
-	buf := prim.NewWriteBuffer(n) // Longest (~132,000); typically < 2^14; RFC https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.2
-	status := fastlyHTTPReqDownstreamTLSClientHello(
-		prim.ToPointer(buf.Char8Pointer()),
-		buf.Cap(),
-		prim.ToPointer(buf.NPointer()),
-	)
-	if status == FastlyStatusBufLen && buf.NValue() > 0 {
-		n = int(buf.NValue())
-		goto alloc // goto saves having to allocate a function closure and avoids having to duplicate the hostcall
+	n := DefaultLargeBufLen // Longest (~132,000); typically < 2^14; RFC https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.2
+	for {
+		buf := prim.NewWriteBuffer(n)
+		status := fastlyHTTPReqDownstreamTLSClientHello(
+			prim.ToPointer(buf.Char8Pointer()),
+			buf.Cap(),
+			prim.ToPointer(buf.NPointer()),
+		)
+		if status == FastlyStatusBufLen && buf.NValue() > 0 {
+			n = int(buf.NValue())
+			continue
+		}
+		if err := status.toError(); err != nil {
+			return nil, err
+		}
+		return buf.AsBytes(), nil
 	}
-	if err := status.toError(); err != nil {
-		return nil, err
-	}
-
-	return buf.AsBytes(), nil
 }
 
 // witx:
@@ -1030,24 +1030,24 @@ func fastlyHTTPReqMethodGet(
 
 // GetMethod returns the HTTP method of the request.
 func (r *HTTPRequest) GetMethod() (string, error) {
-	n := DefaultSmallBufLen
-alloc:
-	buf := prim.NewWriteBuffer(n)
-	status := fastlyHTTPReqMethodGet(
-		r.h,
-		prim.ToPointer(buf.Char8Pointer()),
-		buf.Cap(),
-		prim.ToPointer(buf.NPointer()),
-	)
-	if status == FastlyStatusBufLen && buf.NValue() > 0 {
-		n = int(buf.NValue())
-		goto alloc // goto saves having to allocate a function closure and avoids having to duplicate the hostcall
+	n := DefaultSmallBufLen // HTTP Methods are short: GET, POST, etc.
+	for {
+		buf := prim.NewWriteBuffer(n)
+		status := fastlyHTTPReqMethodGet(
+			r.h,
+			prim.ToPointer(buf.Char8Pointer()),
+			buf.Cap(),
+			prim.ToPointer(buf.NPointer()),
+		)
+		if status == FastlyStatusBufLen && buf.NValue() > 0 {
+			n = int(buf.NValue())
+			continue
+		}
+		if err := status.toError(); err != nil {
+			return "", err
+		}
+		return buf.ToString(), nil
 	}
-	if err := status.toError(); err != nil {
-		return "", err
-	}
-
-	return buf.ToString(), nil
 }
 
 // witx:
@@ -1097,23 +1097,23 @@ func fastlyHTTPReqURIGet(
 // GetURI returns the fully qualified URI of the request.
 func (r *HTTPRequest) GetURI() (string, error) {
 	n := DefaultMediumBufLen // Longest (unknown); Typically less than 1024, but some browsers accept much longer
-alloc:
-	buf := prim.NewWriteBuffer(n)
-	status := fastlyHTTPReqURIGet(
-		r.h,
-		prim.ToPointer(buf.Char8Pointer()),
-		buf.Cap(),
-		prim.ToPointer(buf.NPointer()),
-	)
-	if status == FastlyStatusBufLen && buf.NValue() > 0 {
-		n = int(buf.NValue())
-		goto alloc // goto saves having to allocate a function closure and avoids having to duplicate the hostcall
+	for {
+		buf := prim.NewWriteBuffer(n)
+		status := fastlyHTTPReqURIGet(
+			r.h,
+			prim.ToPointer(buf.Char8Pointer()),
+			buf.Cap(),
+			prim.ToPointer(buf.NPointer()),
+		)
+		if status == FastlyStatusBufLen && buf.NValue() > 0 {
+			n = int(buf.NValue())
+			continue
+		}
+		if err := status.toError(); err != nil {
+			return "", err
+		}
+		return buf.ToString(), nil
 	}
-	if err := status.toError(); err != nil {
-		return "", err
-	}
-
-	return buf.ToString(), nil
 }
 
 // witx:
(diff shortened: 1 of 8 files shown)


internal/abi/fastly/hostcalls_guest.go
```diff
@@ -905,7 +910,7 @@ func fastlyHTTPReqHeaderValuesSet(
 func (r *HTTPRequest) SetHeaderValues(name string, values []string) error {
 	var buf bytes.Buffer
 	for _, value := range values {
-		fmt.Fprint(&buf, value)
+		buf.WriteString(value)
```
joeshaw, line 913: 👍

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -3493,27 +3545,19 @@ func fastlyDeviceDetectionLookup(
 ) FastlyStatus
 
 func DeviceLookup(userAgent string) ([]byte, error) {
-	buf := prim.NewWriteBuffer(defaultBufferLen)
-
 	userAgentBuffer := prim.NewReadBufferFromString(userAgent).Wstring()
-
+	n := DefaultSmallBufLen // Longest JSON of https://www.fastly.com/documentation/reference/vcl/variables/client-request/client-identified/
```
joeshaw, line 3549: I would also lean toward Medium here.

joeshaw:
> Due to the failing device detection test, there's something not working as expected with adaptive buffers.
> 
> Specifically: I thought that `buf.NValue()` would report the size of the buffer needed to hold the full value, if the passed buffer was too short. Instead it always equals `0`.

It's possible the device detection code doesn't support adaptive buffers.  Probably an oversight.

joeshaw:
> > Due to the failing device detection test, there's something not working as expected with adaptive buffers.
> > Specifically: I thought that `buf.NValue()` would report the size of the buffer needed to hold the full value, if the passed buffer was too short. Instead it always equals `0`.
> 
> It's possible the device detection code doesn't support adaptive buffers. Probably an oversight.

Looks like a bug in Viceroy: https://github.com/fastly/Viceroy/blob/main/lib/src/wiggle_abi/device_detection_impl.rs#L46-L51

Note how `n_written` is not set to the expected size of the buffer.

Contrast this with the equivalent secret store code: https://github.com/fastly/Viceroy/blob/main/lib/src/wiggle_abi/secret_store_impl.rs#L105-L114

It doesn't appear to be an issue with the production Compute platform though.

joeshaw:
https://github.com/fastly/Viceroy/pull/383 adds adaptive buffer support for geolocation and device detection to Viceroy, which should allow these tests to pass once it's merged and a new Viceroy is released.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -636,13 +638,19 @@ func fastlyHTTPReqDownstreamTLSClientHello(
...
+	status := fastlyHTTPReqDownstreamTLSClientHello(
 		prim.ToPointer(buf.Char8Pointer()),
 		buf.Cap(),
 		prim.ToPointer(buf.NPointer()),
-	).toError(); err != nil {
+	)
+	if status == FastlyStatusBufLen && buf.NValue() > 0 {
+		n = int(buf.NValue())
+		goto alloc // avoid all the allocations of a new stack, etc.
+	}
```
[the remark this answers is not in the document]
joeshaw: Yeah, I was just thinking of how we do it in other places like https://github.com/fastly/compute-sdk-go/blob/bcb1f0a4f338fcfc56088b3986308f7bc6498c84/internal/abi/fastly/hostcalls_guest.go#L2798-L2815

It reassigns (but doesn't re-declare) `buf` and `status` by calling the functions again without a `goto`.

I might be the one misreading (or misunderstanding)... the comment suggested to me that the `goto` is necessary to avoid allocations.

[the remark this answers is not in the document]
cee-dub: I think the comment just lacks context in that if we did a function closure to call the hostcall again that it would probably undo a good amount of our memory savings.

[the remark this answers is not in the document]
cee-dub: I didn't like repeating the call in a nested conditional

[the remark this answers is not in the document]
joeshaw: Ah, ok!  Understood now, thanks.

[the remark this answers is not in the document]
cee-dub: I've updated these comments to try to clarify.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -636,13 +638,19 @@ func fastlyHTTPReqDownstreamTLSClientHello(
...
 		prim.ToPointer(buf.NPointer()),
-	).toError(); err != nil {
+	)
+	if status == FastlyStatusBufLen && buf.NValue() > 0 {
+		n = int(buf.NValue())
+		goto alloc // goto saves having to allocate a function closure and avoids having to duplicate the hostcall
+	}
+	if err := status.toError(); err != nil {
 		return nil, err
 	}
```
lpereira, line 656: I'm fine with the `goto`, but maybe this could be an infinite loop instead? Then there's no need to explain why you're using it.

```go
for {
     buf := ...
     status := ...
     if status == FastlyStatusBufLen && buf.NValue() > 0 {
             n = int(buf.NValue())
             continue
     }
     if err := ... {
             return nil, err
     }

     return buf.AsBytes(), nil
}
```

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -1084,15 +1095,21 @@ func fastlyHTTPReqURIGet(
...
 		prim.ToPointer(buf.Char8Pointer()),
 		buf.Cap(),
 		prim.ToPointer(buf.NPointer()),
-	).toError(); err != nil {
+	)
+	if status == FastlyStatusBufLen && buf.NValue() > 0 {
+		n = int(buf.NValue())
+		goto alloc // goto saves having to allocate a function closure and avoids having to duplicate the hostcall
+	}
+	if err := status.toError(); err != nil {
```
lpereira, line 1112: All these methods are copied all over, so maybe we could refactor them by providing a function that takes a `func` that actually makes the hostcall, and handles the adaptive buffer handling?

Then these API entrypoints would look something like this:

```go
func (r *HTTPRequest) GetSomething() (string, error) {
     n := DefaultMediumBufLen // explanation why this is the default here
     if buf, err := hostcallWithAdaptiveBufferSize(n, fastlyHTTPReqSomethingGet); err != nil {
           return nil, err
     }
     return buf.ToString(), nil
}
```

lpereira:
I like these changes! From a cursory glance I didn't see anything obviously wrong, just the minor nits I pointed above.

internal/abi/fastly/hostcalls_guest.go
cee-dub: I'd love to do this but the type system will fight us on passing the various hostcalls, since they all have different signatures. To get around that we need a closure like `func() { fastlyHostcall(args, args, args) }` and then I think we start losing the allocation battle.

lpereira: Yeah, you're right.  Tinygo isn't great at figuring out closures are essentially forwarding arguments to another function and generating something better. :(

internal/abi/fastly/hostcalls_guest.go
cee-dub: Here's an attempt for `Dictionary#GetBytes()` (it only partially support adaptive buffers with a known maximum size):
``` go
// Get the value for key, as a byte slice, if it exists.
func (d *Dictionary) GetBytes(key string) ([]byte, error) {
	keyBuffer := prim.NewReadBufferFromString(key).Wstring()
	n := DefaultMediumBufLen // Longest (8192) = Config Store limit; typical values likely less than 1024
	for {
		buf := prim.NewWriteBuffer(n)
		status := fastlyDictionaryGet(
			d.h,
			keyBuffer.Data, keyBuffer.Len,
			prim.ToPointer(buf.Char8Pointer()), buf.Cap(),
			prim.ToPointer(buf.NPointer()),
		)
		if status == FastlyStatusBufLen && n < dictionaryValueMaxLen {
			// The Dictionary API cannot return the needed size with this error.
			// Instead of perfectly adapting, we allocate the maximum length a value can have.
			n = dictionaryValueMaxLen
			continue
		}
		if err := status.toError(); err != nil {
			return nil, err
		}
		return buf.AsBytes(), nil
	}
}
```

cee-dub: Hard to say if this produces more allocations with the `:=` in the loop than the goto version. I'm not even sure how to test that.

lpereira: It shouldn't be different! `for {}` is just a `goto` and a label in a trenchcoat.

lpereira: The original writebuf allocation will be lost if an reallocation was needed, so maybe we could also have a buffer pool to reuse allocations made with adaptive buffer sizes that had to be reallocated?

cee-dub: Yep, pool will come in a following PR

elliottt:
I've verified on the ExecuteD side that all the host calls (except for `fastly_dictionary#get`) will populate the length needed outparam. Additionally, the two Viceroy host calls that you mentioned being unsure about (`fastly_cache#get_user_metadata` and `fastly_http_req#downstream_tls_client_hello`) are both currently unimplemented.

internal/abi/fastly/hostcalls_guest.go
cee-dub: Thanks for the comments. Improved the end result!

PR #119, opened 2024-06-14
Clarify the behavior of `Pass` in `CacheOptions`

acfoltzer:
In helping a user troubleshoot the caching behavior of their application, I noticed that it's easy to get the wrong idea from the current documentation that setting `Pass` to `false` prevents a request from ever bypassing the cache. This is not the case; the `Pass` option only has an effect when `true`, and its effect is to force the cache to be bypassed completely. I have edited the docstrings to hopefully make that more clear.


cee-dub (approved):
Thanks, this is much clearer.

PR #120, opened 2024-06-18
abi/fastly: correct max value length and use scratch buffer in Dictionary#GetBytes

cee-dub:
This PR updates work from #117 to handle [documented limits](https://docs.fastly.com/en/guides/about-edge-dictionaries#limitations-and-considerations) for config stores correctly.

In #117 a heuristic was applied to buffers to start small, and increase to a size above the documented limit of "8000 characters." Those characters, however, are UTF-8 encoded, and thus could be comprised of as many as 8000 * 3 bytes. This PR correctly adjusts up to that maximum limit, with an intermediate step for up to 8000 UTF-8 encoded ASCII-range characters, which need 8000 bytes.

It also introduces a scratch space buffer for copies across the host/guest boundary and passes up copies of returned values from the config store.


integration_tests/config_store/main_test.go
```diff
@@ -60,4 +61,25 @@ func TestConfigStore(t *testing.T) {
...
+	if got != want {
+		t.Errorf("Body = %q, want %q", got, want)
+	}
+
+	maxUTF8, err := d.Get(strings.Repeat("ゝ", 255))
```
lpereira, line 76: Haven't seen this character in a long time! :)

lpereira (approved):

integration_tests/config_store/main_test.go
cee-dub: Had to find a good one that was 3 bytes in UTF-8 and the repetition mark seemed like a good fit here. 🍺

fgsch:
Have we confirmed both key and values are utf-8 encoded? 
I recall at least keys being limited to a small subset of ascii characters, so utf-8 wouldn't apply.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2480,26 +2484,17 @@ func fastlyDictionaryGet(
...
+	status := fastlyDictionaryGet(
+		d.h,
+		keyBuffer.Data, keyBuffer.Len,
+		prim.ToPointer(buf.Char8Pointer()), buf.Cap(),
+		prim.ToPointer(buf.NPointer()),
+	)
+	if err := status.toError(); err != nil {
+		return nil, err
 	}
+	return buf.AsBytes(), nil
```
lpereira, line 2497: Are our tests sufficient to prove that `buf.AsBytes()` is copying out `valueBuf` to a new buffer, rather than optimizing the slicing out as merely a view inside `valueBuf` itself?

I would imagine something like:

```go
config.Set("foo", "123456789")
config.Set("bar", "abcdefghi")

value1, _ := config.GetBytes("foo")
value2, _ := config.GetBytes("bar")

if value1 != "123456778" { ... }
if value2 != "abcdefghi" { ... }
```

cee-dub: You make a good point. I think this is OK as-is:

The slice operation inside `AsBytes` isn't meant to copy, since we're still in `internal` and we control consuming the slice. I wanted to avoid an internal copy to a new `[]byte` before `Get` copies it implicitly with `string()`.

I used the three element form in the shared code so any `append` calls to the result of `AsBytes()` would be forced to make a copy, since that's way more common than directly re-slicing or modifying indices of a `[]byte` in place. This will help protect future additions of shared volatile buffers to other host calls.

The config store code only passes the volatile slice up to the public part of the SDK (to be copied to a new `[]byte`) before end users have access to the data, so perhaps this test is best at that level.

internal/abi/fastly/hostcalls_guest.go
```diff
@@ -2436,8 +2436,13 @@ func fastlyDictionaryOpen(
 // Dictionary represents a Fastly edge dictionary, a collection of read-only
 // key/value pairs. For convenience, keys are modeled as Go strings, and values
 // as byte slices.
+//
+// NOTE: wasm, by definition, is a single-threaded execution environment. This
```
dgryski, line 2440: While wasm is single-threaded, both TinyGo and Big Go have support for goroutines via their own preemption strategies.  While we can probably get away with a single buffer, I still think it should be protected by a mutex.

cee-dub: Completely reasonable

internal/abi/fastly/hostcalls_guest.go
dgryski: This requirement should probably be documented otherwise we're going to forget.  Could we tweak this internal API to make doing the wrong thing harder?

cee-dub: > This requirement should probably be documented otherwise we're going to forget. Could we tweak this internal 
API to make doing the wrong thing harder?

@dgryski I agree, this felt risky. Using a mutex is much cleaner and gives a place to document the expected behavior and subtleties.

@lpereira `config.Set` does not exist. 😮

dgryski (approved):
I think this is fine to merge, and since it doesn't change the public API we're safe to roll-back if we're unhappy with it.

PR #124, opened 2024-06-27
fsttest: discard header changes after send

fgsch:
This should correctly ignore headers set too late to make it into the response.

### DIFF
--- a/fsttest/recorder.go
+++ b/fsttest/recorder.go
@@ -32,6 +32,8 @@ func (r *ResponseRecorder) Header() fsthttp.Header {
 	if !r.headersDone {
 		return r.HeaderMap
 	}
+	// Once the send the headers, return a copy so any changes
+	// are discarded.
 	return r.HeaderMap.Clone()
 }
 
--- a/fsttest/recorder.go
+++ b/fsttest/recorder.go
@@ -12,9 +12,10 @@ import (
 // ResponseRecorder is an implementation of fsthttp.ResponseWriter that
 // records its mutations for later inspection in tests.
 type ResponseRecorder struct {
-	Code      int
-	HeaderMap fsthttp.Header
-	Body      *bytes.Buffer
+	Code        int
+	HeaderMap   fsthttp.Header
+	Body        *bytes.Buffer
+	headersDone bool
 }
 
 // NewRecorder returns an initialized ResponseRecorder.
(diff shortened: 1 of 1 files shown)


fsttest/recorder.go
```diff
@@ -28,12 +29,18 @@ func NewRecorder() *ResponseRecorder {
 
 // Header returns the response headers to mutate within a handler.
 func (r *ResponseRecorder) Header() fsthttp.Header {
-	return r.HeaderMap
+	if !r.headersDone {
+		return r.HeaderMap
+	}
+	return r.HeaderMap.Clone()
```
joeshaw, line 35: Do you mind adding a comment here to explain the purpose?  I'm not sure it's obvious without additional context.

joeshaw (commented):
LGTM but I have a request for a comment explaining why we return a clone.

fsttest/recorder.go
cee-dub: This is a good point and I agree. Being part of our test harness, I was initally being more liberal about the lack of comment.

fgsch: Sure, I will update the PR.

fgsch: Done. Let me know if that works for you.

fsttest/recorder.go
```diff
@@ -28,12 +29,20 @@ func NewRecorder() *ResponseRecorder {
 
 // Header returns the response headers to mutate within a handler.
 func (r *ResponseRecorder) Header() fsthttp.Header {
-	return r.HeaderMap
+	if !r.headersDone {
+		return r.HeaderMap
+	}
+	// Once the send the headers, return a copy so any changes
+	// are discarded.
+	return r.HeaderMap.Clone()
```
cee-dub, line 37:
```suggestion
	if !r.headersSent {
		return r.HeaderMap
	}
	// Once sent, return a copy so any changes are discarded.
	return r.HeaderMap.Clone()
```
I find the name `headersSent` to be more self-explanatory than `headersDone`, so I suggest changing that throughout.

cee-dub (commented):
Could have given better feedback the first time through. Apologies for the back and forth!

fsttest/recorder.go
fgsch: Done.

fgsch:
Updated to address feedback. PTAL

cee-dub (approved):
LGTM!

PR #129, opened 2024-08-15
Add support for the get_vcpu_ms hostcall

acw:
This adds support for a hostcall that fetches the amount of time that the guest has spent on the VCPU, but does not include any time the guest has spent idle. The Viceroy implementation [is forthcoming](https://github.com/fastly/Viceroy/pull/412).

The raw hostcall provides this value in milliseconds, but the little reading on time in Go suggested that translating this to a 
`Duration` would be more natural. I can revisit this decision, or obviously folks can reach in and get the raw millisecond value from the internal library.

The test case included makes sure that we don't count sleeping time. It may be worth adding an additional test to verify that we do count active time.

### DIFF
--- a/compute/compute.go
+++ b/compute/compute.go
@@ -30,7 +30,7 @@ func GetVCPUTime() (time.Duration, error) {
     return 0, err
   }
 
-  var result time.Duration = time.Duration(milliseconds) * time.Millisecond 
+  result := time.Duration(milliseconds) * time.Millisecond 
 
   return result, nil
 }
--- a/compute/compute.go
+++ b/compute/compute.go
@@ -1,7 +1,7 @@
 // Copyright 2024 Fastly, Inc.
 
 // Useful functions for interacting with the compute instance runtime.
-package compute_runtime
+package compute
 
 import (
   "time"
--- a/compute/compute.go
+++ b/compute/compute.go
@@ -30,8 +30,7 @@ func GetVCPUTime() (time.Duration, error) {
     return 0, err
   }
 
-  var result time.Duration
-  result = time.Duration(milliseconds) * time.Millisecond 
+  var result time.Duration = time.Duration(milliseconds) * time.Millisecond 
 
   return result, nil
 }
--- a/compute/compute.go
+++ b/compute/compute.go
@@ -4,9 +4,9 @@
 package compute
 
 import (
-  "time"
+	"time"
 
-  "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
 )
 
 // Get the amount of time taken on the vCPU.
@@ -24,13 +24,13 @@ import (
 // The resulting percentage should be relatively stable across different
 // platforms, and useful in doing A/B testing.
 func GetVCPUTime() (time.Duration, error) {
-  milliseconds, err := fastly.GetVCPUMilliseconds()
+	milliseconds, err := fastly.GetVCPUMilliseconds()
 
-  if err != nil {
-    return 0, err
-  }
+	if err != nil {
+		return 0, err
+	}
 
-  result := time.Duration(milliseconds) * time.Millisecond 
+	result := time.Duration(milliseconds) * time.Millisecond
 
-  return result, nil
+	return result, nil
 }
--- a/compute_runtime/compute_runtime.go
+++ b/compute_runtime/compute_runtime.go
@@ -1,5 +1,6 @@
 // Copyright 2024 Fastly, Inc.
-//
+
+// Useful functions for interacting with the compute instance runtime.
 package compute_runtime
 
 import (
--- a/compute_runtime/compute_runtime.go
+++ b/compute_runtime/compute_runtime.go
@@ -1,0 +1,22 @@
+// Copyright 2024 Fastly, Inc.
+//
+package compute_runtime
+
+import (
+  "time"
+
+  "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+)
+
+func GetVCPUTime() (time.Duration, error) {
+  milliseconds, err := fastly.GetVCPUMilliseconds()
+
+  if err != nil {
+    return 0, err
+  }
+
+  var result time.Duration
+  result = time.Duration(milliseconds) * time.Millisecond 
+
+  return result, nil
+}
--- a/compute_runtime/compute_runtime.go
+++ b/compute_runtime/compute_runtime.go
@@ -9,6 +9,20 @@ import (
   "github.com/fastly/compute-sdk-go/internal/abi/fastly"
 )
 
+// Get the amount of time taken on the vCPU.
+//
+// The resulting time is millisecond-accurate, but we recommend against
+// comparing the absolute values returned across different runs (or builds)
+// of the program.
+//
+// Because compute guests can run on a variety of different platforms,
+// you should not necessarily expect these values to converge across
+// different sessions. Instead, we strongly recommend using this value
+// to look at the relative cost of various operations in your code base,
+// by taking the time before and after a particular operation and then
+// dividing this by the total amount of vCPU time your program takes.
+// The resulting percentage should be relatively stable across different
+// platforms, and useful in doing A/B testing.
 func GetVCPUTime() (time.Duration, error) {
   milliseconds, err := fastly.GetVCPUMilliseconds()
 
--- a/compute_runtime/doc.go
+++ b/compute_runtime/doc.go
@@ -1,0 +1,5 @@
+// Copyright 2024 Fastly, Inc.
+//
+// Package compute_runtime provides ways to reflect on the operation of the
+// compute runtime in which your application is operating.
+package compute_runtime
--- a/compute_runtime/doc.go
+++ b/compute_runtime/doc.go
@@ -1,5 +1,5 @@
 // Copyright 2024 Fastly, Inc.
-//
+
 // Package compute_runtime provides ways to reflect on the operation of the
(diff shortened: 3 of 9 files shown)


compute_runtime/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright 2024 Fastly, Inc.
+//
+// Package compute_runtime provides ways to reflect on the operation of the
+// compute runtime in which your application is operating.
```
joeshaw, line 4: I know this is weird but here and in other places you'll want to separate the copyright header comment from the comment block that applies to the `package` line or the `package` line itself.  Otherwise it will end up in the documentation.

compute_runtime/compute_runtime.go
```diff
@@ -0,0 +1,22 @@
...
+package compute_runtime
+
+import (
+  "time"
+
+  "github.com/fastly/compute-sdk-go/internal/abi/fastly"
+)
+
+func GetVCPUTime() (time.Duration, error) {
+  milliseconds, err := fastly.GetVCPUMilliseconds()
```
joeshaw, line 12: The check failures here are because you will also need to add the hostcalls to `hostcalls_noguest.go`, which is what gets compiled when the `go:build` conditions from your new `compute_runtime_guest.go` file don't apply.  It's fine to just have them return a not-implemented error like the other functions in that file.

compute_runtime/compute_runtime.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2024 Fastly, Inc.
+//
+package compute_runtime
```
joeshaw, line 3: As an idiomatic thing, underscores aren't used much in Go identifiers.  (Much to my chagrin.)  So we should find a different name for this.  Some options:

* `computeruntime` -- this is a lot and my eyes can't help but see "computer" in there
* `compute` - ok
* `runtime` - there's a stdlib package called this, but the likelihood that some code will import both is small.  Still, maybe best to avoid it
* `compute/runtime` - same issue as the stdlib package.  this would nest a level in the filesystem though.

My ranked voting would be `compute` first and `compute/runtime` second.

joeshaw (commented):
The code itself looks good, so my comments below are mostly build and idiom related.

compute_runtime/compute_runtime.go
acw: Oh, interesting. That explains some naming things I've noticed here and there. OK, in that case, I went with `compute`, sort of. I noticed with the "internal" versions, it looks like we are maintaining the underscore and then only removing it for the handy ones we expect folks to use. (Specifically, I note `device_detection` and `secret_store`.) So I left things as `compute_runtime` in `internal`, and then only changed it to `compute` on the "outside".

If those two examples are not two I should've learned from, I'm happy to change the `internal` version too. 
:laughing:

compute_runtime/doc.go
acw: Cool, think I caught these in ff3c1f5

compute_runtime/compute_runtime.go
acw: Thank you for the pointer! Would've taken me forever to find this. Done now.

compute/compute.go
```diff
@@ -0,0 +1,36 @@
...
+// The resulting percentage should be relatively stable across different
+// platforms, and useful in doing A/B testing.
+func GetVCPUTime() (time.Duration, error) {
+  milliseconds, err := fastly.GetVCPUMilliseconds()
+
+  if err != nil {
+    return 0, err
+  }
+
+  var result time.Duration = time.Duration(milliseconds) * time.Millisecond
```
fgsch, line 33:
```suggestion
  result := time.Duration(milliseconds) * time.Millisecond 
```

integration_tests/compute_runtime/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+  now, err := compute.GetVCPUTime()
+  if err != nil {
+    t.Errorf("Couldn't get starting vcpu time")
+  }
+
+  var counter uint64
+  var next time.Duration
+
+  counter = 0
+  next = now
```
fgsch, line 37:
```suggestion
  counter := 0
  next := now
```

integration_tests/compute_runtime/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+  "github.com/fastly/compute-sdk-go/compute"
+)
+
+func TestGetVcpuMs(t *testing.T) {
+  start, err := compute.GetVCPUTime()
+  if err != nil {
+    t.Errorf("Couldn't get starting vcpu time")
+  }
+
+  time.Sleep(5 * time.Second)
```
cee-dub, line 17: This is a really long sleep for tests, especially tests that don't run in parallel. I'd suggest something much shorter, and if you need to burn vCPU time, do some guest code work for a bit.

integration_tests/compute_runtime/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+    t.Errorf("Couldn't get ending vcpu time")
+  }
+
+  if end - start > time.Second {
+    t.Errorf("Sleeping shouldn't count as vcpu time!")
+  }
+
+  now, err := compute.GetVCPUTime()
+  if err != nil {
+    t.Errorf("Couldn't get starting vcpu time")
```
cee-dub, line 30: Message is copied but not updated.

internal/abi/fastly/compute_runtime_guest.go
```diff
@@ -0,0 +1,41 @@
...
+// session.
+//
+// Because compute guests can run on a variety of different platforms,
+// you should not necessarily expect these values to converge across
+// different sessions. Instead, we strongly recommend using this value
+// to look at the relative cost of various operations in your code base,
+// by taking the time before and after a particular operation and then
+// dividing this by the total amount of vCPU time your program takes.
+// The resulting percentage should be relatively stable across different
+// platforms, and useful in doing A/B testing.
```
cee-dub, line 30: Did you mean to duplicate this in the internal package?

cee-dub (changes requested):

compute_runtime/compute_runtime.go
joeshaw: You came to the right conclusion.  Underscores in filenames are fine, and the convention.  But two "stronger" conventions are (1) package names should match directory names and (2) identifiers should not have underscores, so these tend to result in directories not having underscores in them.

If you want to dive deeper into this, see https://go.dev/blog/package-names.

joeshaw:
The test failures here make me wonder if we need to bump our Viceroy requirement?

integration_tests/compute_runtime/main_test.go
dgryski: A plain `0` won't have `uint64` type.  (But I +1 the `next := now` tweak).

dgryski:
The timout during the test makes me think it might be a scheduler issue where the process isn't running asynchronously because something is stalled in the sleep. I can look into this more.

integration_tests/compute_runtime/main_test.go
acw: OK, I think 22cbc93 is the combination of your suggestions?

integration_tests/compute_runtime/main_test.go
acw: It's more a question of trying to reduce possible flakes. Having big delays with big difference checks really reduces the possibility that we'll start seeing CI do some exciting things. In 839b205 I've updated the test down to 2 seconds, and then making sure that we haven't spent half a second in the sleep. I can go further, but my lack of knowledge of the Go runtime is making me worry that a random GC might cause a flake if I take the test too low.

integration_tests/compute_runtime/main_test.go
acw: Yup. Should be fixed in c06a439.

internal/abi/fastly/compute_runtime_guest.go
acw: I wasn't sure how visible this API is to users. Is the internal/external separation more of a style thing, or is it that users won't be able to see/use this?

Because this API is one that's easy to misinterpret, my inclination was to put warning labels everywhere.

acw: Happy to remove it, though, if it's just wasted bytes.

integration_tests/compute_runtime/main_test.go
cee-dub: Go GC pauses are microseconds, in general, but wasi is a new world, so it is hard to know. Do the tests run on big Go, TinyGo, both?

internal/abi/fastly/compute_runtime_guest.go
cee-dub: Users won’t see this, but no harm in leaving it for us developers of the library. I do worry that the two identical comments will drift over time, but that’s not a major concern.

integration_tests/compute_runtime/main_test.go
acw: They should run on anything that runs on Viceroy. And I will admit I may be overly concerned about flakes, I just want to make sure the people who might see the flakes in the future are aware of the possibility. (So I'm happy to dial this down to a second sleep and a check that we haven't use 200 milliseconds. After that I worry about diminishing returns.)

cee-dub: 5x faster for a single test is a nice speed up! Flakes wasting people’s time are also bad, so I agree wholeheartedly to striking an appropriate balance.

acw: Done in d203ed1!

cee-dub (approved):

PR #133, opened 2024-08-20
add client certificate and grpc backend builder options

dgryski:


.github/workflows/integration-tests.yml
```diff
@@ -19,6 +19,9 @@ jobs:
           - go-version: '1.21' # pairs with TinyGo 0.29.0
             tinygo-version: '0.29.0'
             wasmtime-version: '21.0.1' # pairs with TinyGo 0.29.0
+          - go-version: '1.23' # pairs with TinyGo 0.29.0
```
cee-dub, line 22:
```suggestion
          - go-version: '1.23' # pairs with TinyGo 0.33.0
```

fsthttp/backend.go
```diff
@@ -279,6 +280,19 @@ func (b *BackendOptions) SNIHostname(host string) *BackendOptions {
 	return b
 }
 
+// ClientCertificate sets the client certificate to be provided to the server as part of the SSL handshake.
+func (b *BackendOptions) ClientCertificate(certificate string, key secretstore.Secret) *BackendOptions {
+	b.abiOpts.UseSSL(true)
```
joeshaw, line 285: Should other SSL-related things (`CertHostname`, `CACert`, `Ciphers`, `SNIHostname`, `SSLMaxVersion`, `SSLMinVersion`) also imply `UseSSL(true)`?  Feels like either all of them should or none of them should, and we should make it clear in the docs if so.

dgryski: This behaviour was copied from the Rust SDK. 
```
    /// Provide the given client certificate to the server as part of the SSL handshake.
    //
    /// Setting this will enable SSL for the connection as a side effect. Both
    /// the certificate and the key to use should be in standard PEM format;
    /// providing the information in another format will lead to an error. We
    /// suggest that (at least the) key should be held in something like the
    /// Fastly secret store for security, with the handle passed to this function
    /// without unpacking it via [`Secret::plaintext`]; the certificate can
    /// be held in a less secure medium.
    ///
    /// (If it is absolutely necessary to get the key from another source, we
    /// suggest the use of [`Secret::from_bytes`]).
    pub fn provide_client_certificate(
        mut self,
        pem_certificate: impl ToString,
        pem_key: Secret,
    ) -> Self {
        self.use_ssl = true;
        self.client_certificate_info = Some((pem_certificate.to_string(), pem_key));
        self
    }
```

The other SSL-related calls also set `use_ssl = true` in the Rust SDK.  I'll update our implementation as well.

cee-dub (approved):
This LGTM.

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
 
 // SSLMinVersion sets the minimum allowed TLS version on SSL connections to this backend.
```
joeshaw, line 247:
```suggestion
// SSLMinVersion sets the minimum allowed TLS version on SSL connections to this backend.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) SSLMinVersion(min TLSVersion) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.SSLMinVersion(fastly.TLSVersion(min))
 	return b
 }
 
 // SSLMaxVersion sets the maximum allowed TLS version on SSL connections to this backend.
```
joeshaw, line 254:
```suggestion
// SSLMaxVersion sets the maximum allowed TLS version on SSL connections to this backend.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) SSLMaxVersion(max TLSVersion) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.SSLMaxVersion(fastly.TLSVersion(max))
 	return b
 }
 
 // CertHostname sets the hostname that the server certificate should declare.
```
joeshaw, line 261:
```suggestion
// CertHostname sets the hostname that the server certificate should declare.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) CertHostname(host string) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.CertHostname(host)
 	return b
 }
 
 // CACert sets the CA certificate to use when checking the validity of the backend.
```
joeshaw, line 268:
```suggestion
// CACert sets the CA certificate to use when checking the validity of the backend.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) CACert(cert string) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.CACert(cert)
 	return b
 }
 
 // Ciphers sets the list of OpenSSL ciphers to support for connections to this origin.
```
joeshaw, line 275:
```suggestion
// Ciphers sets the list of OpenSSL ciphers to support for connections to this origin.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) Ciphers(ciphers string) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.Ciphers(ciphers)
 	return b
 }
 
 // SNIHostname sets the SNI hostname to use on connections to this backend.
```
joeshaw, line 282:
```suggestion
// SNIHostname sets the SNI hostname to use on connections to this backend.
// Setting this will enable SSL for the connection as a side effect.
```

fsthttp/backend.go
```diff
@@ -245,40 +246,59 @@ func (b *BackendOptions) UseSSL(v bool) *BackendOptions {
...
 func (b *BackendOptions) SNIHostname(host string) *BackendOptions {
+	b.abiOpts.UseSSL(true)
 	b.abiOpts.SNIHostname(host)
 	return b
 }
 
+// ClientCertificate sets the client certificate to be provided to the server as part of the SSL handshake.
```
joeshaw, line 289:
```suggestion
// ClientCertificate sets the client certificate to be provided to the server as part of the SSL handshake.
// Setting this will enable SSL for the connection as a side effect.
```

joeshaw (approved):
Doc suggestions below.  Otherwise LGTM.

fsthttp/backend.go
dgryski: Updated all the doc comments and force-pushed.

PR #134, opened 2024-08-20
bump tinygo version

dgryski:


.github/workflows/validate.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           go-version: stable
```
cee-dub, line 42: We talked about using `oldstable` here to keep from having breakage with new Go releases. Thoughts on bumping it back to that?

.github/workflows/validate.yml
```diff
@@ -57,7 +57,4 @@ jobs:
         run: go test -v -race -tags="fastlyinternaldebug nofastlyhostcalls" ./...
 
       - name: Tests - TinyGo
-        env:
-          GOARCH: wasm
-          GOOS: wasip1
-        run: tinygo test -v -tags="fastlyinternaldebug nofastlyhostcalls" ./...
+        run: tinygo test -v -target=wasip1 -tags="fastlyinternaldebug nofastlyhostcalls" ./...
```
cee-dub, line 60: Personally, I'd love to reduce noise in the test output by dropping `-v` here and elsewhere. Maybe a different PR?

cee-dub (approved):
LGTM. A couple optional things.

.github/workflows/validate.yml
dgryski: Done.

.github/workflows/validate.yml
dgryski: So the assumption is devs will discover breakages on their machines well before CI hits.  Ok.

cee-dub:
Great!

cee-dub (approved):

PR #136, opened 2024-08-28
vendor net url

dgryski:


dkegel-fastly:
fyi sigsci-edge is now building from this branch, and the build WILL BREAK if anything is pushed to this branch (go is picky that way)

dkegel-fastly:
Closed in favor of https://github.com/fastly/compute-sdk-go/pull/141

PR #137, opened 2024-08-30
Expose the ability to turn pooling on or off for dynamic backends.

acw:
This option looks like it might have been skipped, but can be very handy.

### DIFF
--- a/fsthttp/backend.go
+++ b/fsthttp/backend.go
@@ -300,6 +300,12 @@ func (b *BackendOptions) ClientCertificate(certificate string, key secretstore.S
 	return b
 }
 
+// PoolConnections allows users to turn connection pooling on or off for the
+// backend. Pooling allows the Compute platform to reuse connections across
+// multiple sessions, resulting in lower resource use at the server (because it
+// does not need to reperform the TCP handhsake and TLS authentication when the
+// connection is reused). The default is to pool connections. Set this to false
+// to create a new connection to the backend for every incoming session.
 func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
 	b.abiOpts.PoolConnections(poolingOn)
 	return b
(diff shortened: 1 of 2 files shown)


dkegel-fastly:
I suspect we need to also add that method on type BackendOptions, because its guts are otherwise private?

Otherwise trying to access it gets 
```
./dynamic_backend.go:264:32: reqIn.DynamicBackend.Options.abiOpts undefined (type *fsthttp.BackendOptions has no field or method abiOpts)

type BackendOptions struct {
        abiOpts fastly.BackendConfigOptions
}
```

internal/abi/fastly/types.go
```diff
@@ -884,6 +884,14 @@ func (b *BackendConfigOptions) SNIHostname(sniHostname string) {
 	b.opts.sniHostnameLen = prim.U32(buf.Len())
 }
 
+func (b *BackendConfigOptions) PoolConnections(poolingOn bool) {
+	if poolingOn {
+		b.mask &^= backendConfigOptionsMaskDontPool
```
dgryski, line 889: It's a pity the ABI defined the flag in the negative :(

fsthttp/backend.go
```diff
@@ -300,6 +300,11 @@ func (b *BackendOptions) ClientCertificate(certificate string, key secretstore.S
 	return b
 }
 
+func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
```
dgryski, line 303: Need a doc comment here.

dgryski (changes requested):

internal/abi/fastly/types.go
acw: This was intentional, so that folks using older SDKs would default to using pooling, which most people want.

fsthttp/backend.go
acw: Done in 373de08!

dgryski (approved):
LGTM

PR #139, opened 2024-09-04
fsthttp, internal: add get_addr_dest_ip/port hostcalls

### LINKED ISSUE #121: Implement two hostcalls for obtaining address/port used to connect to origin
Two hostcalls were recently added. Support for them has already been added in the Rust SDK, but should also be added in the Go SDK.

dgryski:
Fixes #121

### DIFF
--- a/fsthttp/response.go
+++ b/fsthttp/response.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
 	"sync"
 
 	"github.com/fastly/compute-sdk-go/internal/abi/fastly"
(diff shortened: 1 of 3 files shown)


dgryski:
Not sure I like how these hostcalls are exposed, but the hostcalls themselves work.

ydnar:
Can you add a `RemoteAddr()` method to return a `net.Addr` instead of exposing host and port as fields on `Response`?

cee-dub:
Agree with other comments that this should be a method to return a friendly standard Go type.

joeshaw:
Consider exposing this as a `net/netip.AddrPort` instead of or in addition to a `net.Addr`

dgryski:
API fixed. PTAL.

cee-dub (approved):
LGTM

PR #140, opened 2024-09-05
Expose the ability to set HTTP and TCP keepalives on dynamic backends.

acw:


fsthttp/backend.go
```diff
@@ -311,6 +311,49 @@ func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
...
+	if interval < time.Second {
+		interval = time.Second
+	}
+
+	b.abiOpts.TcpKeepaliveInterval(interval)
+
+	return b
+}
+
+// TcpKeepaliveProbes sets how many failed TCP probes we should send to the
```
ulyssa, line 341: Folks will probably understand either way, but I find "unanswered" clearer:

```suggestion
// TcpKeepaliveProbes sets how many unanswered TCP probes we should send to the
```

fsthttp/backend.go
```diff
@@ -311,6 +311,49 @@ func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
...
+// TcpKeepaliveEnable sets whether or not to use TCP keepalives to try to
+// maintain the connetion to the backend.
+func (b *BackendOptions) TcpKeepaliveEnable(enable bool) *BackendOptions {
+	b.abiOpts.TcpKeepaliveEnable(enable)
+	return b
+}
+
+// TcpKeepaliveInterval sets the interval to use when sending TCP keepalive
+// probes. Setting this value implicitly enables TCP keepalives. Intervals of
+// less than 1 second will be rounded up to 1 second.
```
ulyssa, line 330: I don't know if it's better to just document it like this or to make the order irrelevant, but it feels like this could end up mattering:

```suggestion
// TcpKeepaliveInterval sets the interval to use when sending TCP keepalive
// probes. Intervals of less than 1 second will be rounded up to 1 second.
//
// Setting this value implicitly enables TCP keepalives. If you are calling both
// this method and `TcpKeepAliveEnable` with dynamically loaded or generated
// values, make sure to call `TcpKeepAliveEnable` last.
```

fsthttp/backend.go
```diff
@@ -311,6 +311,52 @@ func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
 	return b
 }
 
+// HttpKeepaliveTime configures how long to allow HTTP connections to remain
+// idle in a connection pool before it should be considered closed.
+func (b *BackendOptions) HttpKeepaliveTime(time time.Duration) *BackendOptions {
```
dgryski, line 316: Do we want `HTTP` instead of `Http` ?

fsthttp/backend.go
```diff
@@ -311,6 +311,52 @@ func (b *BackendOptions) PoolConnections(poolingOn bool) *BackendOptions {
...
+	b.abiOpts.HttpKeepaliveTime(time)
+	return b
+}
+
+// TcpKeepaliveEnable sets whether or not to use TCP keepalives to try to
+// maintain the connetion to the backend.
+func (b *BackendOptions) TcpKeepaliveEnable(enable bool) *BackendOptions {
```
dgryski, line 323: Naming: `TCP` vs `Tcp`

dgryski (changes requested):
Naming tweaks.

fsthttp/backend.go
dkegel-fastly: https://pkg.go.dev/net/http#ParseHTTPVersion says yes...?

fsthttp/backend.go
dkegel-fastly: random hit says yes: https://pkg.go.dev/github.com/m-lab/tcp-info/tcp uses TCP

fsthttp/backend.go
acw: Fixed in 107ec31!

fsthttp/backend.go
acw: Fixed in 107ec31!

internal/abi/fastly/types.go
```diff
@@ -913,6 +928,40 @@ func (b *BackendConfigOptions) UseGRPC(v bool) {
...
+}
+
+func (b *BackendConfigOptions) TCPKeepaliveInterval(t time.Duration) {
+	b.mask |= backendConfigOptionsMaskKeepalive
+	b.opts.tcpKeepaliveEnable = prim.U32(1)
+	b.opts.tcpKeepaliveIntervalSecs = prim.U32(t.Seconds())
+}
+
+func (b *BackendConfigOptions) TCPKeepaliveProbes(count uint32) {
+	if count > 0 {
```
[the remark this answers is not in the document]
acw: I have a little bit of hesitation about using 0 to unset (I just don't like special values like that), but yes, that's a possibility. Basically, using 0 here would return this value to being the default for the system. If you wanted to turn probes off, you'd still need to unset KeepaliveEnable.

PR #141, opened 2024-09-12
fsthttp: SetParseRequestURI now lets you register a url parser

dkegel-fastly:
This is like https://github.com/fastly/compute-sdk-go/pull/136 but more suitable for polite company, with less bloat, less opinion about the right way to parse urls, and near zero overhead unless opted in.

The new entry point is behind its own little build tag, fastlyinternalsetparseuri, because it may be needed at different times than the things behind our other private api build tag.

Tested with a real app.

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -133,6 +133,9 @@ func NewRequest(method string, uri string, body io.Reader) (*Request, error) {
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

cee-dub (commented):
LGTM. Thanks for providing a hook.

PR #143, opened 2024-09-25
add secretstore.Plaintext convenience function

joeshaw:
This consolidates opening the store, getting the secret, and decrypting
the secret into a single call, making more a more convenient API for
getting secrets.

Update the integration test and example to use this, since it does
all the same stuff as before.

### DIFF
--- a/secretstore/secretstore.go
+++ b/secretstore/secretstore.go
@@ -129,3 +129,22 @@ func SecretFromBytes(b []byte) (*Secret, error) {
 
 	return &Secret{s: s}, nil
 }
+
+// Plaintext decrypts and returns the secret value as a byte slice for
+// the given secret store and secret name.
+//
+// This is a convenience function that combines the functionality of
+// [Open], [Store.Get], and [Secret.Plaintext].
+func Plaintext(storeName, secretName string) ([]byte, error) {
+	st, err := Open(storeName)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


secretstore/secretstore.go
```diff
@@ -129,3 +129,22 @@ func SecretFromBytes(b []byte) (*Secret, error) {
 
 	return &Secret{s: s}, nil
 }
+
+// Plaintext decrypts and returns the secret value as a byte slice for
+// the given secret store and secret name.
+//
+// This is a convenience function that combines the functionality of
+// [Open], [Store.Get], and [Secret.Plaintext].
```
dgryski, line 137: Do we want to add a caveat if that if you're obtaining multiple secrets from a store to actually call `Open/Get/Plaintext` etc instead of calling this repeatedly?

joeshaw: IMO we can skip it, but I'll add it if you feel strongly about it.  There's very little overhead, basically just the `open` hostcall itself and copying a handle or error back into guest space.

cee-dub: Also, 🚲 🏠 ⚠️ but I prefer phrasing it as `returns the secret value as a []byte for`

dgryski (approved):
LGTM.  We can tweak doc wording later.

PR #144, opened 2024-10-11
cache/core: update docs for request headers in transactions

joeshaw:
Attempting to set request headers in write options for
Transaction.Insert or Transaction.InsertAndStreamBack returns
ErrInvalidArgument.  The docs previously only stated this for
Transaction.Update.

### DIFF
--- a/cache/core/core.go
+++ b/cache/core/core.go
@@ -326,7 +326,8 @@ type WriteOptions struct {
 	// RequestHeaders are a set of HTTP headers that influence cache
 	// lookups when Vary rules are set.
 	//
-	// This field is only valid for Insert.  If provided for Update,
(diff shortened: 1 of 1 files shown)


dgryski:
Do you have a link to the Rust SDK docs or other source for this change?

joeshaw:
The best I have here is https://docs.rs/fastly/latest/fastly/cache/core/struct.TransactionInsertBuilder.html#method.vary_by but it's not an exact mapping because the Rust SDK uses the builder pattern and the Go SDK uses an options struct.

This was found by a customer report, and then I found the error handling for this on the insert handler in the Compute platform's code.  I can follow up with you privately with the links to those.

dgryski (approved):
LGTM

PR #146, opened 2024-11-15
acl: add ACL hostcalls

dgryski:


dgryski:
Note we can't merge this or run any tests until we have a viceroy release including https://github.com/fastly/Viceroy/pull/438

acl/acl.go
```diff
@@ -0,0 +1,115 @@
...
+	// Is it a acl-specific error?
+	if aclErr, ok := err.(fastly.ACLError); ok {
+		switch aclErr {
+		case fastly.ACLErrorUninitialized: // we really shouldn't be returning this
+			return ErrUnexpected
+		case fastly.ACLErrorOK:
+			// Not an error; we shouldn't get here
+			return fmt.Errorf("%w (%s)", ErrUnexpected, err)
+		case fastly.ACLErrorNoContent:
+			return ErrNoContent
```
awilliams-fastly, line 95: In the context of `Handle.Lookup`, I wonder if something like `ErrNoMatch` would be more easily understood?

dgryski: So replace `ErrNoContent` with a new error `ErrNoMatch` which improves the semenatic match but pulls it farther away from the underlying response code (and Rust SDK I guess).

awilliams-fastly: Yes, agree that it pulls it further from the underlying response code. The Rust SDK uses `Option<Response>` with `None` when there's no match. 

I think `ErrNoContent` is fine, but wanted to ask in case we could think of anything better.

dgryski: The other way to represent "not found" would be to return a pointer to the entry, or a nil pointer (and no error) if it's not found.   That puts the onus on the user, but they're already checking for error (and in fact maybe this is "nicer" since it doesn't conflate an error and a success-not-found case.)

awilliams-fastly: Yes, I think using a pointer/nil could work and agree that it's not conflating errors with not-found cases. One final alternative would be to return something like:
`(r Result, ok bool, err error)`
 where ok is false in the not-found case.

dgryski: Rethinking this; I think I'm going to stick with returning an error.  For example, Go's `database/sql` package returns `ErrNoRows` for query returning no rows.  This is the same.   And I want to stick with the error name in the wit file for consistency (and also in case things change in the future we won't end up in a situation where our SDK code drifts *more* from the underlying host calls.)

awilliams-fastly (approved):

_examples/acl/main.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"net"
+
+	"github.com/fastly/compute-sdk-go/acl"
+	"github.com/fastly/compute-sdk-go/fsthttp"
+)
+
+func main() {
+	fsthttp.ServeFunc(func(ctx context.Context, w fsthttp.ResponseWriter, r *fsthttp.Request) {
+		aclh, err := acl.Open("example")
```
awilliams-fastly, line 16: Understanding this is an example, should the `err` be handled?

awilliams-fastly (approved):

_examples/acl/main.go
dgryski: Oops, thanks.

dgryski: Fixed and force-pushed.

awilliams-fastly (approved):

PR #150, opened 2025-01-28
internal: remove deprecated reflect.SliceHeader

cee-dub:
Replaces the deprecated `reflect.SliceHeader` with the recommended `unsafe.Slice` and `unsafe.SliceData`, as appropriate.

### DIFF
--- a/internal/abi/prim/prim.go
+++ b/internal/abi/prim/prim.go
@@ -3,7 +3,6 @@
 package prim
 
 import (
-	"reflect"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

cee-dub:
An app is deployed using this branch of the SDK here: https://possibly-supreme-deer.edgecompute.app/

> [!NOTE]
> The response is slow to appear in the browser since this uses `_examples/parallel-requests` which receives slow responses from the `httpme` backend.

PR #165, opened 2025-05-21
Reject out-of-order ranges

cceckman-at-fastly:
Reject ranges where the start position ("from") comes after the end position ("to").

These don't have a good definition of how they should behave; reject them.


dgryski (approved):
LGTM.   Thanks.

PR #166, opened 2025-05-30
Handle config store values larger than 8k

awilliams-fastly:
The maximum allowed length for a config store entry's value is now configurable. The default is still 8k characters, but may be larger. This changes the Config Store implementation to re-size its buffer when encountering a BUFLEN error and try again.


dgryski (approved):
LGTM.  Integration tests pass.

PR #169, opened 2025-06-12
Option to respect range during cache streaming

cceckman-at-fastly:
The core cache API allows for cache items to be concurrently streamed to / from the cache.

If:
- The size of the cached item's body was not provided by the writer
- The reader requests a specific range of the cached item's body
- The writer and reader are concurrent, i.e. the body is streamed from one to the other

then today, the core cache API will ignore the requested range and provide the whole body. There is no explicit notification that the whole body is provided instead of a range.

In this SDK release, this remains the default behavior. However, the lookup options struct now now offers an `AlwaysUseRequestedRange` option. If set to true, body reads conducted as part of this lookup/transaction will always use the requested range, even when streaming. If false, the behavior is unchanged (as above).

In a future SDK release, the default behavior will change to
"always use requested range".

### DIFF
--- a/cache/core/core.go
+++ b/cache/core/core.go
@@ -258,6 +258,16 @@ type LookupOptions struct {
 	// To indicates the ending offset to read from the cached object.  A
 	// value of 0 means to read to the end of the object.
 	To uint64
+
+	// By default, if:
+	// - The size of the cached item's body was not provided by the writer
+	// - The reader requests a specific range of the cached item's body (`From` and `To` are set)
+	// - The writer and reader are concurrent, i.e. the body is streamed from one to the other
+	// then today, the core cache API will ignore the requested range and provide the whole body.
+	//
+	// Setting this flag provides the more intuitive behavior: the range will be respected
+	// during streaming as well.
+	AlwaysUseRequestedRange bool
 }
 
 func abiLookupOptions(opts LookupOptions) (fastly.CacheLookupOptions, error) {
@@ -271,6 +281,8 @@ func abiLookupOptions(opts LookupOptions) (fastly.CacheLookupOptions, error) {
 
 		abiOpts.SetRequest(req)
 	}
+
+	abiOpts.SetAlwaysUseRequestedRange(opts.AlwaysUseRequestedRange)
 
 	return abiOpts, nil
 }
--- a/cache/core/core_test.go
+++ b/cache/core/core_test.go
@@ -208,3 +208,46 @@ func ExampleFound_GetRange() {
 
 	fmt.Printf("The cached value was: %s", cachedStr)
 }
+
+func ExampleLookupOptions_AlwaysUseRequestedRange() {
+	const (
+		key      = "my_key"
+		contents = "my cached object"
+	)
+
+	// Start an insert; this will be concurrent with the read.
+	// Data written to this handle is streamed into the Fastly cache.
+	w, err := core.Insert([]byte(key), core.WriteOptions{
+		TTL:           time.Hour,
+		SurrogateKeys: []string{key},
+		Length:        uint64(len(contents)),
+	})
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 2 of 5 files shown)


cache/core/core.go
```diff
@@ -258,6 +258,16 @@ type LookupOptions struct {
 	// To indicates the ending offset to read from the cached object.  A
 	// value of 0 means to read to the end of the object.
 	To uint64
+
+	// By default, if:
```
dgryski, line 262: Go comments should always start with the name of the symbol being described, with later paragraphs providing more details.

```
AlwaysUseRequestedRange forces the provided range to be used during streaming.

If false, under certain circumstances the the entire body will be returned instead of just the requested range.
<extra details here>
```

Although, if the issue is sufficiently rare *and* users will "always" want the sane behaviour, maybe we should invert this bool and fix the issue for users in those rare cases, and have them toggle the flag to have the old behaviour.  (But really would someone *want* the old behaviour?  Or would it just manifest as a weird bug for them?)

dgryski (changes requested):
Requesting a change for at least the comment formatting, but also to open a discussion about the API in general.

cache/core/core.go
cceckman-at-fastly: Ack on comments, I'll fix that.

In terms of "API direction" - yeah, we do want to (at some point) switch the default. But, storytime: my original test for this change _hung_ when I "just" switched the default. The check of "is the range valid" happens at the "get the body" point, i.e. the `get_body` hostcall is blocking. In my test, the whole body was written after the `get_body` call - so the `get_body` call deadlocked.

I think it's not _likely_ that a customer would run into this -- if the HTTP client is making a range request, they probably want ETag validation, not a range from a different body. But it's _possible_, so I didn't want to change the default without marking the relevant release as breaking.

My plan for Rust is to make this ~same change in 0.11.5, and cut a 0.12 release in the not-too-distant future with some other major-version-motivating changes. For Go - I'm happy to go with whatever you suggest. :)

cache/core/core_test.go
```diff
@@ -208,3 +208,46 @@ func ExampleFound_GetRange() {
...
+	if _, err := io.WriteString(w, contents); err != nil {
+		panic(err)
+	}
+
+	cachedStr, err := io.ReadAll(f.Body)
+	if err != nil {
+		panic(err)
+	}
+	if string(cachedStr) != "cached" {
+		panic(fmt.Sprintf("got: %q, want: %q", cachedStr, "cached object"))
```
cee-dub, line 249:
```suggestion
		panic(fmt.Sprintf("got: %q, want: %q", cachedStr, "cached"))
```

cache/core/core_test.go
```diff
@@ -208,3 +208,46 @@ func ExampleFound_GetRange() {
...
+	}
+
+	fmt.Printf("The cached value was: %s", cachedStr)
```
cee-dub, line 252: This should be removed before merging. If you want this to always appear in test logs (in verbose mode or when the test fails, use t.Logf:

```suggestion
	t.Logf("The cached value was: %s", cachedStr)
```

cee-dub (commented):
A couple of minor things that should be addressed before merge.

dgryski:
Failing CI

cache/core/core_test.go
cceckman-at-fastly: I accepted this, but this caused the CI failure - this is an Example, not a Test, so there's no `t`.

cee-dub: You're right! Apologies. I missed that when looking at (and writing code in the Github UI of) the PR.

cache/core/core.go
cceckman-at-fastly: Switched the default in ebe1351

cceckman-at-fastly:
Done as a squash on the PR, with an updated message reflecting the behavior.

PR #172, opened 2025-08-08
all: add support for shielding hostcalls

dgryski:


shielding/shielding.go
```diff
@@ -0,0 +1,54 @@
...
+		target:    info.Target(),
+		sslTarget: info.SSLTarget(),
+	}, nil
+}
+
+// Name returns the name of the shield site.
+func (s *Shield) Name() string { return s.name }
+
+// Me returns whether the Compute node is currently in the shielding site.
+func (s *Shield) Me() bool { return s.me }
```
cceckman-at-fastly, line 34: Two cents: I find "running on" the clearer name for this, but your call.

shielding/shielding.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Target returns the target for unecrypted data.
+func (s *Shield) Target() string { return s.target }
+
+// SSLTarget returns the target for encrypted traffic.
+func (s *Shield) SSLTarget() string { return s.sslTarget }
```
cceckman-at-fastly, line 40: At some point someone did a pass through the Rust SDK with `s/ssl/tls/`; I guess that was before adding the shielding code (I see it says "SSL" in the Rust side too).

shielding/shielding.go
```diff
@@ -0,0 +1,54 @@
...
+
+// BackendOptions
+type BackendOptions struct {
+	cacheKey string
+}
```
cceckman-at-fastly, line 45: For what it's worth, we don't expose the `BackendOptions` (or even the target URLs) in the Rust SDK today. Maybe that's setting us up for a breaking change in the future if we add new options :/ so your call on including it in public.

I would probably advise against exposing `cacheKey`, though- I haven't seen a clear explanation of the intended semantics (if you have, point me to it!) and it's a no-op today- so if we were to start using it for something, that would be a change in XQD that alters behavior for an unmodified guest. Easier if we don't expose it in the SDKs to begin with, then if we add a use for it, we can expose it in the SDK.

cceckman-at-fastly (changes requested):
Hiding `cacheKey` is the only thing that I feel strongly about, others are just nits / opinions.

dgryski:
PTAL

shielding/shielding.go
```diff
@@ -0,0 +1,40 @@
...
+
+// IsRunningOn returns whether the Compute node is currently in the shielding site.
+func (s *Shield) IsRunningOn() bool { return s.runningOn }
+
+// BackendOptions
+type BackendOptions struct {
+}
+
+// Backend returns a named backend for use with the fsthttp package.
+func (s *Shield) Backend(opts *BackendOptions) (string, error) {
```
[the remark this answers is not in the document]
dgryski: Yes.  There was an (unexported) option in the Rust SDK that I copied over, but have since removed.

dgryski:
Last minute fixup; changed the info target parsing code but at the same time removed the API calls that were checking they worked.  (Spoiler: it didn't..)

cee-dub:
Force pushes make it hard to tell what changed (I'd be fine with a squashed merge at the end for this OSS repo, though!)

dgryski:
Yeah, I realized afterwards that this kind of fixup probably should have been a separate commit and then squashed at merge time.
Basically there was an off-by-one in the new code because I changed my mind about stripping the initial zero or not.

```
diff --git a/internal/abi/fastly/shielding_guest.go b/internal/abi/fastly/shielding_guest.go
index a7ab34d..9103242 100644
--- a/internal/abi/fastly/shielding_guest.go
+++ b/internal/abi/fastly/shielding_guest.go
@@ -58,7 +58,7 @@ func ShieldingShieldInfo(name string) (*ShieldInfo, error) {
        info.me = bufb[0] == 1

        if !info.me {
-               // extract targets
+               // strip first null byte and extract targets
                vals := bytes.Split(bufb[1:], []byte{0})

                // ensure we got what we expected
@@ -66,8 +66,8 @@ func ShieldingShieldInfo(name string) (*ShieldInfo, error) {
                        return nil, FastlyStatusBadAlign.toError()
                }

-               info.target = string(vals[1])
-               info.sslTarget = string(vals[2])
+               info.target = string(vals[0])
+               info.sslTarget = string(vals[1])
        }
```

PR #176, opened 2025-08-14
fsthttp: warn if WriteHeader is called multiple times

### LINKED ISSUE #164: Flag multiple calls to WriteHeader()
Much like the Go `net/http` package catches these and warns, our `fsthttp` package should similarly warn when `WriteHeader` is called multiple times.

dgryski:
Fixes #164


joeshaw (approved):
LGTM 👍

PR #179, opened 2025-08-20
Add reusable sessions

dgryski:


internal/abi/fastly/types.go
```diff
@@ -1672,3 +1686,43 @@ func (s *ShieldInfo) RunningOn() bool { return s.me }
...
+//     (flags (@witx repr u32)
+//         $reserved
+//         $timeout
+//     ))
+
+type nextRequestOptionsMask prim.U32
+
+const (
+	nextRequestOptionsMaskReserved nextRequestOptionsMask = 0b0000_0001 // $reserved
+	nextRequestOptionsMaskTimeout  nextRequestOptionsMask = 0b0000_0010 // $request_headers
```
joeshaw, line 1710:
```suggestion
	nextRequestOptionsMaskTimeout  nextRequestOptionsMask = 0b0000_0010 // $timeout
```

fsthttp/handle.go
```diff
@@ -20,23 +16,75 @@ var (
...
+	if err != nil {
+		panic(fmt.Errorf("create client ResponseWriter: %w", err))
+	}
+
+	h.ServeHTTP(ctx, clientResponseWriter, clientRequest)
+	clientResponseWriter.Close()
+}
+
+// ServeMany is sugar for Serve(HandlerFunc(f)).
+func ServeMany(h HandlerFunc, nextTimeout time.Duration, maxRequests int, maxLifetime time.Duration) {
```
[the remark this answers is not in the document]
dgryski: Good point.  I initially only had one argument, and then added more during development.  I'll make this an options struct.

internal/abi/fastly/hostcalls_noguest.go
ulyssa, line 1: I don't know if you'll want to do it as part of this change or another one, but there's a `fastly_http_downstream::downstream_client_request_id` hostcall that returns a per-request UUID, which will be helpful for customers wanting to uniquely identify each incoming request in some way.

joeshaw: Is this the same as `FASTLY_TRACE_ID`?

dgryski: Yes, my initial work was focused on getting the parts of `fastly_http_downstream` I needed for reusable sessions.  Later this week I will expose the rest of the hostcalls from that module.

ulyssa: > Is this the same as FASTLY_TRACE_ID?

When there's a single request, the two are identical. But the value of `FASTLY_TRACE_ID` remains the same throughout the duration of the session, so it's more of a "session UUID" than a "request UUID", especially since it's what is used in the `fastly log-tail` output for correlating all logs from the same session.

joeshaw (commented):
Does Viceroy have support for reusable sessions?  Is it possible to also add an integration test for this?

joeshaw (approved):
LGTM 👍

dgryski:
(Turns out we don't run our integration tests through the normal viceroy Serve infrastructure -- can't write a test here.)

PR #189, opened 2025-09-16
fix(fsthttp): restore previous limit method return values

cee-dub:


cee-dub:
It appears we need to update another integration test since `underlyingReaderFrom()` can detect the reader inside an `io.NopCloser`.

joeshaw (approved):
👍

cee-dub:
> Can you say more?  When I last looked it was not possible to see inside of `io.NopCloser`, which is why we have our own implementation of it in `fsthttp` which can.

I may have misread the code in seeking an explanation for the repeated failure (flake?) of the `io.NopCloser` test that's currently failing in CI.

PR #198, opened 2025-09-26
Add support for image optimization

dgryski:


fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,787 @@
...
+const (
+	EnableOptUpscale EnableOpt = "upscale"
+)
+
+func (e EnableOpt) isSet() bool    { return e != "" }
+func (e EnableOpt) String() string { return string(e) }
+
+type Opts struct {
+	Region                             Region
+	PreserveQueryStringOnOriginRequest bool
```
cee-dub, line 535: As a blank line so the short names below line up with less indentation.

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,787 @@
...
+	Quality                            uint32
+	ResizeFilter                       ResizeAlgorithm
+	Saturation                         int
+	Sharpen                            *Sharpen
+	Trim                               *Sides
+	TrimColor                          *TrimColor
+	Width                              PixelsOrPercentage
+}
+
+func encodeCommas(s string) string {
```
cee-dub, line 566:
```suggestion
func encodeCommasAndColons(s string) string {
```

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,787 @@
...
+			return err
+
+		}
+	}
+
+	return nil
+}
+
+func (o *Opts) QueryString() string {
+	var args []string
```
cee-dub, line 676: url.Query is no good here?

fsthttp/imageopto/imageopto.go
dgryski: Removed in the end.  I only added this so I could use the Rust test cases unmodified, but imageopto doesn't need them to be encoded.

fsthttp/imageopto/imageopto.go
dgryski: Wanted to preserve ordering to make the test cases easier. (But yes technically I could.  Don't think it would be nicer though)

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,964 @@
...
+
+func (p PixelsOrPercentage) validate() error {
+	if !rePixelsOrPercentage.MatchString(string(p)) {
+		return fmt.Errorf("imageopto: pixels or percentage: invalid: %q", string(p))
+	}
+
+	return nil
+}
+
+var rePixelsOrPercentage = regexp.MustCompile("^[0-9]+(.[0-9]+)?p?$")
```
cee-dub, line 82:
```suggestion
var rePixelsOrPercentage = regexp.MustCompile("^[0-9]+(\.[0-9]+)?p?$")
```

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,964 @@
...
+
+	if p.X.isSet() {
+		return p.X.String()
+	}
+
+	return p.Y.String()
+}
+
+var reXPosition = regexp.MustCompile("^(x[0-9]+p?)|(offset-x[0-9]+)$")
+var reYPosition = regexp.MustCompile("^(y[0-9]+p?)|(offset-y[0-9]+)$")
```
cee-dub, line 200: There's a literal 'p' in the non-offset version?

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,964 @@
...
+}
+
+func (area *Area) String() string {
+	switch area.state {
+	case areaStateNone:
+		return ""
+	case areaStateAspectRatio:
+		return fmt.Sprintf("%v:%v", area.aspect.w, area.aspect.h)
+	case areaStateWidthHeight:
+		return fmt.Sprintf("%v,%v", area.widthHeight.w.String(), area.widthHeight.h.String())
```
cee-dub, line 163: I think you could change all of these to use the `Stringer` implicitly, even leaving the `%v` (vs. changing to `%s`), e.g.

```suggestion
		return fmt.Sprintf("%v,%v", area.widthHeight.w, area.widthHeight.h)
```

fsthttp/imageopto/imageopto.go
```diff
@@ -0,0 +1,964 @@
...
+	// dimension larger if needed.
+	FitCover Fit = "cover"
+	// FitCrop resizes and crops the image centrally to exactly fit the specified region.
+	FitCrop Fit = "crop"
+)
+
+func (f Fit) isSet() bool    { return f != "" }
+func (f Fit) String() string { return string(f) }
+
+// Level specifies a set of constraints indicating a degree of required decoder performance] for a profile.
```
cee-dub, line 515:
```suggestion
// Level specifies a set of constraints indicating a degree of required decoder performance for a profile.
```

dgryski:
PTAL.  Squashed and ready to merge.  Docs copied from the Rust SDK.

cee-dub (approved):
Also, FWIW, [url.Values.Encode](https://pkg.go.dev/net/url#Values.Encode) does sort by key.

fsthttp/imageopto/imageopto.go
dgryski: `x and y can be set as a value in pixels (e.g., 40 is 40 pixels) or as a percentage suffixed with p (e.g., 50p is 50%).`

dgryski: Hmm... I wonder if that needs to have a possible decimal for the percentage too.

cee-dub: OK, cool, I just didn't see that in the example valid values comment.

fsthttp/imageopto/imageopto.go
dgryski: I think I'll merge as is and we can remove `fmt` later.  I'll file an issue.

fsthttp/request.go
```diff
@@ -115,6 +116,8 @@ type Request struct {
 	// RequestID is the current Fastly request ID
 	RequestID string
 
+	ImageOptimizerOpts *imageopto.Opts
```
ccampbell, line 119: Do you think it may make sense to name this `ImageOptimizerOptions`? I ask because this struct already has `CacheOptions` and `DecompressResponseOptions`.

The Rust SDK uses `ImageOptimizerOptions`, and that was also what it was named in the rough design doc we shared.

dgryski: Yeah, good point.  I'll handle this.

dgryski: Sorry, took a while to get to this: https://github.com/fastly/compute-sdk-go/pull/208

PR #201, opened 2025-10-03
fsthttp: support sending StatusEarlyHints

dgryski:


dgryski:
Viceroy support will be in 0.16.  Hopefully will release soon, but this is otherwise unblocked.

https://github.com/fastly/Viceroy/pull/553

PR #206, opened 2025-10-23
clean up terminology around SendError

joeshaw:
Talk about specific error values rather than error types.  These are all
of the same type.


cee-dub (approved):
Nice.

PR #208, opened 2025-10-29
fsthttp: ImageOptimizerOpts -> ImageOptimizerOptions

dgryski:


ccampbell:
> Only question here is if I should rename the type in the `imageopto` package too.

I suppose if I had to pick, I probably prefer `imageopto.Options`, but I leave it up to you

cee-dub (approved):

PR #215, opened 2025-11-14
add SandboxRequests to fastlyMeta

joeshaw:
This lets handlers track how many requests have been handled
by the current sandbox.


cee-dub (approved):
LGTM

PR #218, opened 2025-11-19
handle invalid 1xx status codes in client responses

joeshaw:
handle invalid 1xx status codes in client responses

Compute does not allow 1xx status codes (except 103 Early Hints) and
returns FastlyStatus::Inval from SendDownstream if it is set.  We don't
check that call for errors and that causes us to silently return an
empty 200 OK response.

Because WriteHeader is infallible, we now capture the error from
SendDownstream on the responseWriter and return it from subsequent
Write, Append, and Close calls.  For the specific case of an invalid
status code, we translate it into a new exported ErrInvalidStatusCode
error.

Because calling SendDownstream consumes the underlying ABI response and
body even in failure, this situation is unrecoverable.  serve() checks
for that SendDownstream error and panics, ensuring that clients will
receive a 500 status code rather than a 200.

The docs for WriteHeader have been updated to explain that 1xx codes
are not permitted, and describe the error behavior.

### DIFF
--- a/end_to_end_tests/loopback/main_test.go
+++ b/end_to_end_tests/loopback/main_test.go
@@ -62,3 +62,26 @@ func doLoopbackRequest(t *testing.T, req *fsthttp.Request) *fsthttp.Response {
 	}
 	return resp
 }
+
+func Test1xxStatusCode(t *testing.T) {
+	req, err := fsthttp.NewRequest("GET", "http://anyplace.horse?status_code=101", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp, err := req.Send(context.Background(), "self")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer resp.Body.Close()
+
+	if got, want := resp.StatusCode, 500; got != want {
+		// Unlike Compute, Viceroy returns a 101 status code until
+		// https://github.com/fastly/Viceroy/pull/557 lands, probably in
+		// v0.16.1.
+		if got == 101 {
+			t.Logf("StatusCode = %d, want: %d; accepting until Viceroy is fixed", got, want)
+		} else {
+			t.Errorf("StatusCode = %d, want: %d", got, want)
+		}
+	}
+}
--- a/fsthttp/response.go
+++ b/fsthttp/response.go
@@ -209,7 +209,11 @@ type ResponseWriter interface {
 	// headers collected by Header. If WriteHeader is not called explicitly,
 	// the first call to Write or Close will trigger an implicit call to
 	// WriteHeader with a code of 200. After the first call to WriteHeader,
-	// subsequent calls have no effect.
+	// subsequent calls will print a warning but have no effect.
+	//
+	// 1xx status codes other than 103 (Early Hints) are not permitted and will
+	// result in an ErrInvalidStatusCode error when calling Write, Close, or
+	// Append, and clients will receive a 500 (Internal Server Error) response.
 	WriteHeader(code int)
 
 	// Write the data to the connection as part of an HTTP reply.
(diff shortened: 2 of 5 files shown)


fsthttp/response.go
```diff
@@ -300,7 +301,29 @@ func (resp *responseWriter) WriteHeader(code int) {
...
+		resp.abiResp, err = fastly.NewHTTPResponse()
+		if err != nil {
+			panic(fmt.Errorf("create response: %w", err))
+		}
+		resp.abiBody, err = fastly.NewHTTPBody()
+		if err != nil {
+			panic(fmt.Errorf("create body: %w", err))
+		}
+		return
+	}
```
joeshaw: I'm a little unsure about this approach, and would like feedback.

Normally, calling WriteHeaders more than once is not allowed.  In fact, we recently added a warning if you do call it more than once in the normal case.  This would make it so you're allowed to call it again, but only in certain circumstances (which are hard to identify) if a subsequent Write call fails?

Replacing the abi response and body is weird.  The Rust SDK just panics in this case.  We could do the same and make it the responsibility of the caller to ensure they're passing a valid status code _before_ calling WriteHeaders.

Another alternative is to not panic but just return errors from Write and Close and make them non-recoverable.  In that case we would always send a blank 500 Internal Server Error, as if a panic had happened, but the rest of the user's code would run.  That would be good if they're doing logging or metrics.

end_to_end_tests/loopback/main_test.go
```diff
@@ -62,3 +62,60 @@ func doLoopbackRequest(t *testing.T, req *fsthttp.Request) *fsthttp.Response {
...
+			t.Fatal(err)
+		}
+		defer resp.Body.Close()
+
+		if got, want := resp.StatusCode, 500; got != want {
+			// Unlike Compute, Viceroy returns a 101 status code until
+			// https://github.com/fastly/Viceroy/pull/557 lands, probably in
+			// v0.16.1.
+			if got == 101 {
+				t.Logf("StatusCode = %d, want: %d; accepting until Viceroy is fixed", got, want)
```
cee-dub: Kinda gross, but I really appreciate the clear reason in the comment.

fsthttp/response.go
```diff
@@ -300,7 +301,29 @@ func (resp *responseWriter) WriteHeader(code int) {
 		}
 		resp.abiResp.SetHeaderValues(key, resp.header.Values(key))
 	}
-	resp.abiResp.SendDownstream(resp.abiBody, true)
+
+	// WriteHeader is infallible, so if we're unable to create the downstream response capture it
+	// and return it on Write and Close calls.  This gives callers the opportunity to call
```
cee-dub:
```suggestion
	// WriteHeader is infallible, so if we're unable to create the downstream response capture the error
	// and return it on Write and Close calls.  This gives callers the opportunity to call
```

fsthttp/response.go
cee-dub: One thing I don't like about the option in the implementation in this PR is it needs to make a call to the host to receive a failure, vs. knowing _a priori_ that these 1xx status codes are disallowed and just dealing with it.

Another thing I don't like much is replacing the response and its body.

fsthttp/response.go
```diff
@@ -310,10 +333,18 @@ func (resp *responseWriter) WriteHeader(code int) {
...
 }
 
-// ErrClosed is returned when attempting to write to a ResponseWriter whose network connection has been closed.
-var ErrClosed = errors.New("connection has been closed")
+var (
+	// ErrClosed is returned when attempting to write to a ResponseWriter whose network connection has been closed.
+	ErrClosed = errors.New("connection has been closed")
+
+	// ErrInvalidStatusCode is returned when attempting to write a resposne with an invalid HTTP status code.
+	ErrInvalidStatusCode = errors.New("invalid HTTP status code")
```
cee-dub: We know pretty well in advance if this needs to be returned.

PR #243, opened 2026-03-31
feat: Adding support for ResVPNProxy data

clintjedwards:
Adding a lookup function for ResVPNProxy. This enables VPN/Proxy intelligence data for specific IP addresses.

Most of this code is copied from other implementations for features, I figured we could start with an extra call off of fsthttp and then if we needed later we could make a standalone package as well just like `geo`.

### DIFF
--- a/internal/abi/fastly/http_guest.go
+++ b/internal/abi/fastly/http_guest.go
@@ -1832,6 +1832,254 @@ func (r *HTTPRequest) DownstreamFastlyKeyIsValid() (bool, error) {
 	return valid.b, nil
 }
 
+//go:wasmimport fastly_http_downstream downstream_resvpnproxy_is_anonymous
+//go:noescape
+func fastlyHTTPDownstreamResVPNProxyIsAnonymous(
+	req requestHandle,
+	result prim.Pointer[bool],
+) FastlyStatus
+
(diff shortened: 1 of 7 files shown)


internal/abi/fastly/http_guest.go
```diff
@@ -1832,6 +1832,254 @@ func (r *HTTPRequest) DownstreamFastlyKeyIsValid() (bool, error) {
...
+}
+
+//go:wasmimport fastly_http_downstream downstream_resvpnproxy_is_vpn_datacenter
+//go:noescape
+func fastlyHTTPDownstreamResvpnproxyIsVpnDatacenter(
+	req requestHandle,
+	result prim.Pointer[bool],
+) FastlyStatus
+
+func (r *HTTPRequest) DownstreamResvpnproxyIsVpnDatacenter() (bool, error) {
```
dgryski: Vpn -> VPN

dgryski (changes requested):

dgryski (approved):
LGTM

dgryski (changes requested):
Oops, forgot I had a few changes needed: the alignment struct and the http constant.

dgryski (approved):
LGTM.  Thanks!

PR #245, opened 2026-04-28
Add tests for backend health

dgryski:


integration_tests/backend/main_test.go
```diff
@@ -99,3 +99,95 @@ func TestDynamicBackend(t *testing.T) {
...
+			_, err = req.Send(ctx, b.Name())
+			if err == nil {
+				t.Errorf("Send had nil error")
+				fsthttp.Error(w, "nil error", fsthttp.StatusInternalServerError)
+				return
+			} else {
+				fmt.Fprintf(w, "error from unhealthy send: %v", err)
+			}
+
+		*/
```
cee-dub, line 173: Did it not work to test that the request fails?

cee-dub (approved):
Seems valuable even without a positive test that unhealthy backend always fails.

integration_tests/backend/main_test.go
dgryski: Correct.  Viceroy will happily send to a backend marked "unhealthy".

PR #246, opened 2026-04-29
add IfGenerationMatch to kvstore.InsertConfig

yyyoichi:
I noticed that `if-generation-match` is supported internally but not exposed in the public `kvstore.InsertConfig`. Since it's a small change (just a few lines), I've opened this PR. This is my first contribution here—please let me know if I missed any conventions!

### DIFF
--- a/integration_tests/kvstore/main_test.go
+++ b/integration_tests/kvstore/main_test.go
@@ -59,6 +59,33 @@ func TestKVStore(t *testing.T) {
 		t.Error("expected Lookup failure after delete")
 	}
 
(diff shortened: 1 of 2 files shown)


dgryski:
Can you add a unit test here? https://github.com/fastly/compute-sdk-go/blob/main/integration_tests/kvstore/main_test.go

yyyoichi:
@dgryski Added a unit test for IfGenerationMatch. I used the C++ SDK's test cases as a reference.

dgryski (approved):
LGTM

PR #247, opened 2026-05-06
resvpnproxy: Add Available field

clintjedwards:
This adds an "Available" field to the response back from resvpnproxy.

Previously I left this field out due to time constraints and because I was unconvinced that it was the best way to communicate to the user the fact that the dataset might not be enabled and that is why the rest of the fields have defaults. I had considered it being a separate, informative error instead might be clearer to the user and can be easily ignored as well.

I remain unconvinced, but since `Available` is part of the spec and docs that decision point has flown.

Tested on https://resvpnproxy-compute-lookup.edgecompute.app

```
~❯ curl -s "https://resvpnproxy-compute-lookup.edgecompute.app/resvpnproxy" | jq
{
  "client_ip": "1.1.1.1",
  "available": true,
  "is_anonymous": false,
  "is_anonymous_vpn": false,
  "is_hosting_provider": false,
  "is_proxy_over_vpn": false,
  "is_public_proxy": false,
  "is_relay_proxy": false,
  "is_residential_proxy": false,
  "is_smart_dns_proxy": false,
  "is_tor_exit_node": false,
  "is_vpn_datacenter": false,
  "vpn_service_name": ""
}
```

### DIFF
--- a/_examples/resvpnproxy/main.go
+++ b/_examples/resvpnproxy/main.go
@@ -41,6 +41,14 @@ func main() {
 			return
 		}
 
+		// Check to see if we have live data by checking Available. It doesn't make sense ingesting the rest of the
+		// response if available is false.
+		if !vpnData.Available {
+			w.WriteHeader(fsthttp.StatusOK)
+			fmt.Fprintf(w, `{"error": "ResVPNProxy dataset not enabled for this service"}`)
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


deg4uss3r (approved):

dgryski (approved):
LGTM

PR #252, opened 2026-05-11
fix: add support for if-generation-match

remore:
Added missing IfGenerationMatch field in InsertConfig struct to fully utilize kv store v2 hostcall capabilities


dgryski:
Duplicate of https://github.com/fastly/compute-sdk-go/pull/246

PR #256, opened 2026-05-21
Support `nil` BackendOptions param in RegisterDynamicBackend

awilliams-fastly:
This function already had a nil check for `options`, so it seems like the intent was to support nil. This covers the other case where it needs to be checked for nil before accessing.


dgryski (approved):
LGTM

Issue #199, opened 2025-10-02
Reduce `fmt` use in `fsthttp/imageopto`


dgryski:
Much of the string concatenation could be optimized to just use `+` instead.

PR #259, opened 2026-05-22
Reduce fmt usage in String() methods

case-fastly:
This addresses #199.

Also found a small bug in there: 

> When `nearest3` rounds to a whole number (fract == 0), the function falls into the integer branch. But that branch formats `int32(f)` — the original unrounded input — not `int32(nearest3)`.

Here's an example: https://go.dev/play/p/OahudI8zyfH


dgryski:
The `nearest3` code is a direct translation of the Rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=878c4313bfa509636e7d46941e99c987

We should fix that too.

dgryski (approved):
LGTM.   Only comment would be if we wanted to call `.String()` on a few of types instead of using `string(x)` so we don't break that encapsulation?

dgryski (approved):
LGTM

PR #262, opened 2026-06-09
feat: add MaxAge and TTLRemaining to CandidateResponse

### LINKED ISSUE #250: Clarify http cache TTL field
The Rust SDK has deprecated `CandidateResponse.TTL` and added `TTLRemaining` and`MaxAge`.

mvanhorn:
## Summary

Adds `MaxAge()` and `TTLRemaining()` methods to `fsthttp.CandidateResponse` and deprecates the ambiguous `TTL()` method, bringing the Go SDK in line with the Rust SDK's cache-TTL surface.

## Why this matters (#250)

`CandidateResponse.TTL()` reports `maxAge - age` (the remaining freshness), but its name does not make clear whether it returns the full freshness lifetime of the cached object or the freshness left right now. The Rust SDK already deprecated its `TTL` accessor and split the concept into `MaxAge` ("how long the object stays fresh") and `TTLRemaining` ("how much freshness is left right now"). This change mirrors that surface so callers can distinguish the two.

## Changes

- `MaxAge() (uint32, error)` — returns the full suggested freshness lifetime (`opts.maxAge`, or the set TTL when `SetTTL`/`CacheOptions.TTL` is in effect).
- `TTLRemaining() (uint32, error)` — returns the remaining freshness, `maxAge - age`, and saturates to `0` for already-stale candidates instead of wrapping the unsigned subtraction.
- `TTL()` is retained as a thin wrapper over `TTLRemaining()` and carries a `// Deprecated:` doc comment so existing callers keep working.
- All three read from the existing `getSuggestedCacheWriteOptions()` result, so there is no ABI change.

## Testing

- `go test ./fsthttp/...` passes.
- `go vet ./...` and `gofmt -l` are clean on the changed files.
- `integration_tests/httpcache` builds for `GOOS=wasip1 GOARCH=wasm`.
- `testAfterSendCandidateResponsePropertiesUncached` now asserts `MaxAge()` returns the full suggested lifetime (3600) and `TTLRemaining()` returns `maxAge - age`, alongside the existing `TTL()` backward-compatibility assertion.

Fixes #250

### DIFF
--- a/fsthttp/cache.go
+++ b/fsthttp/cache.go
@@ -378,11 +378,13 @@ func (candidateResponse *CandidateResponse) Age() (uint32, error) {
 	return opts.age, nil
 }
 
-// TTL returns the Time to Live (TTL) in seconds in the cache for this response.
+// MaxAge returns the duration of "freshness", in seconds, for the cached
+// response after it is inserted into the cache.
 //
-// The TTL determines the duration of "freshness" for the cached response
-// after it is inserted into the cache.
-func (candidateResponse *CandidateResponse) TTL() (uint32, error) {
+// This is the full freshness lifetime of the object: it does not subtract the
+// time the object has already spent in cache (its age). Use TTLRemaining to
+// find out how much freshness is left right now.
+func (candidateResponse *CandidateResponse) MaxAge() (uint32, error) {
 	if candidateResponse.useTTL {
 		return candidateResponse.overrideTTL, nil
 	}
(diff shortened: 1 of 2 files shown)


fsthttp/cache.go
```diff
@@ -391,7 +393,52 @@ func (candidateResponse *CandidateResponse) TTL() (uint32, error) {
...
+// TTLRemaining returns the remaining "freshness", in seconds, of the cached
+// response: how much longer the object stays fresh from now.
+//
+// This is the object's full freshness lifetime (MaxAge) minus the time it has
+// already spent in cache (its age). When the object is already stale (its age
+// meets or exceeds its full lifetime) the remaining freshness is reported as 0
+// rather than wrapping around.
+func (candidateResponse *CandidateResponse) TTLRemaining() (uint32, error) {
+	var maxAge, age uint32
+	if candidateResponse.useTTL {
```
dgryski, line 412: I think I'd prefer this if block to look like
```
		opts, err := candidateResponse.getSuggestedCacheWriteOptions()
		if err != nil {
			return 0, err
		}
		maxAge, age = opts.maxAge, opts.age
        if candidateResponse.useTTL {
           maxAge = candidateResponse.overrideTTL
        }
```

dgryski (changes requested):
Minor suggested tweak, but otherwise LGTM.

Thanks!

fsthttp/cache.go
mvanhorn: Done in 00b3a4c. fsthttp tests pass locally.

dgryski (approved):
LGTM!

mvanhorn:
Appreciate you landing this, @dgryski. Good to get MaxAge and TTLRemaining in and the ambiguous TTL deprecated.

PR #265, opened 2026-06-24
fsthttp: ensure SIE flag is passed to ABI opts

dgryski:


posborne (approved):
Fix confirmed in out-of-tree test that initially reproduced the problem; patch makes sense.

PR #268, opened 2026-06-30
improve error logic in examples and tests

dgryski:


kashish-fastly:
I reviewed the PR diff. It addresses the fail-open concern.
It also improves many of the raw-error-to-client examples by logging details server-side and returning generic status text, which is the right pattern.
PR 268 does not change fsthttp/cookie.go, so the Cookie.Domain logging issue remains separate and should be treated as inherited/low-value stdlib-parity behavior unless we have evidence that sensitive or user-controlled values reach Cookie.Domain.

There are still raw error responses in _examples/http-adapter/main.go via w.Write([]byte(err.Error())), and in integration_tests/dynamic_backend/main_test.go via fsthttp.Error(w, err.Error(), ...). -- You should handle them separately

dgryski:
I've removed the ones from the http-adapter example and the backend test.  PTAL.

PR #271, opened 2026-07-14
integration_tests/request_upstream: make sure response coming back is…

dgryski:
… for request we sent


dgryski:
Spend the afternoon debugging the io.NopCloser flake.  As far as I can tell, our code is doing the right thing.  So either something with upstream or something else in the way.  :/

joeshaw (approved):
👍

PR #274, opened 2026-07-27
internal/abi/fastly: http_status is a u16

dgryski:


dgryski:
```
internal/abi/fastly/http_guest.go:3176:6: go:wasmimport: unsupported parameter type httpStatus
```

Huh.
https://github.com/golang/go/blob/14015be5bb4ef69c710ac5a4fad5a5d2a548e3fb/src/cmd/compile/internal/ssagen/abi.go#L369

I feel like this might bite us eventually.

PR #275, opened 2026-07-28
Return descriptive error when fanout is not enabled

serra-fastly:
# Summary
Adds ErrFanoutNotEnabled to surface a descriptive error when HandoffFanout() or HandoffWebsocket() is called on a service without fanout enabled.

Checks FastlyStatus::Unsupported via IsFastlyError() in both call sites in the fsthttp package.

# Context
Previously these calls returned a generic error which was confusing for consumers. Now users get a clear message: "Fanout or WebSockets are not enabled on this service; please contact support for help."

### DIFF
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -1233,9 +1233,10 @@ func (r *Request) HandoffFanout(backend string) error {
 		return ErrHandoffNotSupported
 	}
 	err := r.downstream.req.HandoffFanout(backend)
-	status, ok := fastly.IsFastlyError(err)
-	if ok && status == fastly.FastlyStatusUnsupported {
-		return ErrFanoutNotEnabled
+	if err != nil {
+		if status, ok := fastly.IsFastlyError(err); ok && status == fastly.FastlyStatusUnsupported {
+			return ErrFanoutNotEnabled
+		}
 	}
 	return err
 }
--- a/fsthttp/request.go
+++ b/fsthttp/request.go
@@ -1197,6 +1197,8 @@ type DecompressResponseOptions struct {
 
 var ErrHandoffNotSupported = errors.New("handoff not supported on this request")
 
+var ErrFanoutNotEnabled = errors.New("Fanout or WebSockets are not enabled on this service; please contact support for help")
+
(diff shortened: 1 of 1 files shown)


fsthttp/request.go
```diff
@@ -1209,7 +1211,12 @@ func (r *Request) HandoffWebsocket(backend string) error {
 	if r.downstream.req == nil {
 		return ErrHandoffNotSupported
 	}
-	return r.downstream.req.HandoffWebsocket(backend)
+	err := r.downstream.req.HandoffWebsocket(backend)
```
dgryski, line 1214: Other places where we similarly remap fastly errors look more like
```
	if err != nil {
		if status, ok := fastly.IsFastlyError(err); ok && status == fastly.FastlyStatusInval {
			return nil, ErrBackendNotFound
		}
```
with the nil check first and using a one-line if for ok/status assignment and check.

fsthttp/request.go
```diff
@@ -1224,7 +1231,12 @@ func (r *Request) HandoffFanout(backend string) error {
 	if r.downstream.req == nil {
 		return ErrHandoffNotSupported
 	}
-	return r.downstream.req.HandoffFanout(backend)
+	err := r.downstream.req.HandoffFanout(backend)
+	status, ok := fastly.IsFastlyError(err)
```
dgryski: Same as above.

dgryski (changes requested):
Code tweaks to match the rest of the repo.

fsthttp/request.go
serra-fastly: fixed! [c2374b3](https://github.com/fastly/compute-sdk-go/pull/275/commits/c2374b349d9322a8f6f358bef3c0050076675d5d)

fsthttp/request.go
serra-fastly: [0711cae](https://github.com/fastly/compute-sdk-go/pull/275/commits/0711caeb506c1d20a9876fa14c9782d4ed11e67f)

dgryski (approved):
LGTM

PR #278, opened 2026-08-10
internal: reduce excess []byte -> string -> []byte conversions

### LINKED ISSUE #17: Remove excess string->[]byte->string conversions
There are places in the SDK that induce unnecessary memory allocations / copying, and thus additional work for the garbage collector.  These could be removed for improved performance.

dgryski:
Draft for now.  There might be more.

Fixes #17


cee-dub (approved):
This appears to use unsafe correctly (as far as I understand it).

PR #279, opened 2026-08-10
fsthttp: allow excluding net/http if not required

### LINKED ISSUE #31: fsthttp/{transport,adapter}.go imports net/http, breaking native mac build?
An app that imports fsthttp, and which builds in both go, tinygo native, and tinygo wasm
(with -tag nofastlyhostcalls) stopped building on tinygo native when we upgraded it to
v0.1.2; wrapping fsthttp/{transport,adapter}.go in another build tag seems to resolve the problem.

So adding a build tag to disable the net/http dependency seems like a good thing to add...
maybe named fastly.compute.nostdhttp for clarity.

dgryski:
This also drops heap usage for a hello world from 524KB to 394KB (tinygo) and from 5.2MB to 4.2MB on Big Go.

Not sure if we should actually update the build tag to exclude by default.   Maybe exclusion is opt-in for now, but when we break the API (wasip3 / wasi-http ) we can make users opt-in to needing net/http support?

Update: this also drops request latency on for a hello world on viceroy from 2-3ms to ~1ms or less for binaries built with tinygo.  No latency difference for Big Go binaries.

Fixes #31

### DIFF
--- a/fsthttp/adapter.go
+++ b/fsthttp/adapter.go
@@ -1,3 +1,5 @@
+//go:build !fastlynonethttp
+
 package fsthttp
 
 import (
--- a/fsthttp/adapter_test.go
+++ b/fsthttp/adapter_test.go
@@ -1,3 +1,5 @@
+//go:build !fastlynonethttp
+
 // This test file is in its own test package to avoid a circular
 // dependency between fsthttp and fsttest.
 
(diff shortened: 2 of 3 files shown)


cee-dub:
This is pretty cool for a pure-fsthttp build.

PR #286, opened 2026-08-20
dynamic backend healthchecks

dgryski:


camilee03 (approved):
Not super familiar with go, but the logic lgtm! Noting as well decided to move the functionality to experimental in the [rust sdk](https://github.com/fastly/fst-compute-sdk-rs/pull/188), not sure if similar logic needs to be applied here.

dgryski:
We don't have the same ability in Go to have a separate "experimental" package that is integrated within an existing type.  I can try to come up with something, but most likely I should tag this with a loud comment saying "This is experimental".

PR #299, opened 2026-09-01
Fix early hints

dgryski:


bahamat (approved):
This looks reasonable to me, with the caveat that I don't know Go very well.

PR #304, opened 2026-09-10
expose async versions of kvstore operations

dgryski:


dgryski:
Not sure I like this.  It's an ugly API and doesn't match anything else we provide in the SDK.

cee-dub (commented):
This doesn't look too hard to rewrite using channels in a similar way to e.g. `context.Context`.

PR #308, opened 2026-09-17
fsthttp,internal: support timeouts for http cache lookups

dgryski:


zkat (approved):

fsthttp/cache.go
```diff
@@ -159,7 +165,44 @@ const (
...
+			}
+
+			interval := cacheLookupSelectTimeout
+			if remaining < interval {
+				interval = remaining
+			}
+			intervalMs := interval.Milliseconds()
+			if intervalMs < 1 {
+				intervalMs = 1
+			}
```
cee-dub: 🚲🏠: I think this can be
```suggestion
			intervalMs := max(1, min(remaining, cacheLookupSelectTimeout).Milliseconds())
```

fsthttp/cache.go
```diff
@@ -159,7 +165,44 @@ const (
...
+
+			ready, err := fastly.HTTPCacheAwaitReady(c, uint32(intervalMs))
```
cee-dub: IIUC, this loop spins calling fastly.HTTPCacheAwaitReady, asking it to wait up to `cacheLookupSelectTimeout` (100ms) or `remaining` clamped to min of 1ms.

dgryski: Correct.

fsthttp/cache.go
dgryski: Yes, probably, but I have to think hard about your version :D

cee-dub: I thought about breaking it into 2 lines, which is a little easier.

cee-dub:
```
			intervalMs := min(remaining, cacheLookupSelectTimeout).Milliseconds()
			ready, err := fastly.HTTPCacheAwaitReady(c, uint32(max(1, intervalMs))) // don't pass 0, that blocks until ready.
```

dgryski: Added in 60a00c2

cee-dub (approved):

PR #311, opened 2026-09-21
examples: add grpc-proxy example

tombojer:
I wanted to proxy my gRPC backend using compute.

Because I haven't found any example for my use case, maybe someone benefits from it. 
If it does not fit just close my PR :) 

I tested it locally with my gRPC backend and cli using unary and server streaming requests.


dgryski (approved):
LGTM, Thanks!

